Issue #6, opened 2016-05-11
Making compileJSONCallback work


axic:
**Intro for anyone trying to understand the context:**

Basically `compileJSONCallback` uses the same interface as `compileJSONMulti` (which supports multiple input files defined in a JSON) with the addition of a `readFileCallback` parameter.

This callback is executed when a missing import statement is encountered. It receives the import path and should return the destination file contents. To be more precise it expects an array with two fields: `contents` and `error`.

Internally this is [wrapped to match](https://github.com/ethereum/solc-js/blob/master/index.js#L22) the `readFileCallback(path, contents, error)` method ([see in C](https://github.com/ethereum/solidity/blob/develop/solc/jsonCompiler.cpp#L49)). It has an Emscripten-wrapping, therefore all parameters are pointers to the Emscripten heap. ([relevant Emscripten doc](http://kripken.github.io/emscripten-site/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#interacting-with-code-call-function-pointers-from-c))

Now in C world we have a [condition](https://github.com/ethereum/solidity/blob/develop/solc/jsonCompiler.cpp#L142) to display an error if both `contents` and `error` are null (i.e. JS didn't touched them).  Otherwise they are copied around appropriately.

**Problem:**

The wrapped callback is executed as evidenced by a `console.log` within, but no matter what is done, upon returning from this callback nothing happens. 

If the callback doesn't touches the `contents`/`error`, they should remain null and the error as mentioned above should be displayed.

Sample code to test:
```
var cb = soljson.Runtime.addFunction(function() { console.log('We should trigger the error') });
var output = compileInternal(input, optimize, cb);
console.log(output);
```

```
bash-3.2$ node test.js 
We should trigger the error
bash-3.2$ 
```

The only way I can get a different result is by setting the pointer to a wrong location: `soljson.setValue(ptr, ptr, '*')`:

```
/Users/alex/Projects/solc/solc-js/soljson.js:1
[...]

PR #7, opened 2016-05-11
Fix the file import callback

axic:
This was such an annoying bug with such a simple resolution. For more details see #6.


axic:
@chriseth I think the field `contents` should be renamed to `content` as it is a single file content

chriseth:
This is embarassing! Thanks for the fix :-)

Issue #4, opened 2016-05-09
Export linker


axic:
The library linking option is not available in Javascript. Would be very useful to have it in the form of:

`solc.linkBytecode(bytecode, <class name / address pairs>)`
`solc.link(<output JSON from compile>, <class name to link>, <class name / address pairs>)`

This could also obsolete the linking code in [browser-solidity](https://github.com/chriseth/browser-solidity/blob/gh-pages/assets/js/universal-dapp.js#L462) (and perhaps in [truffle](https://github.com/ConsenSys/truffle/blob/master/lib/contracts.js#L353)).

It could be
a) implemented in Javascript in `solc-js` or
b) `solidity/solc/jsonCompiler.cpp` could be extended to export `link()` too (as it is present internally).

chriseth:
This concerns more solidity in general, but I would like to bring this up here:

Solidity does not have something like a "global namespace" on "globally unique namespace hierarchy". This could be a good thing, at least it is again a parallel to javascript (at least npm and I think also es6) where imports only modify the current scope. When seen from "the outside", this creates problems: There is no way to uniquely refer to a contract or library by name, you always have to specify the scope.

If there are two libraries with the same name, how do we uniquely distinguish them for the sakes of the linker? `<filename>/libraryname`?

axic:
Good point.  Not totally sold whether the `<filename>` part should be the a) caller's filename (the file where the class referencing it is) or b) a reference to the import directive.  In the case of option b) a separator other than `/` might be more appropriate.

chriseth:
With `<filename>` I meant the (potentially absolute) path of the file that contains the library definition.

redsquirrel:
Not sure if it makes sense to bring it up here, but one way to resolve naming collisions would be to somehow delegate that conflict to one or more of the emerging Solidity package managers, who would ensure unique names within their registries.

axic:
It is related to import statements too because the library might be defined via a URL (see your Github URL parser in browser-solidity)

tcoulter:
> If there are two libraries with the same name, how do we uniquely distinguish them for the sakes of the linker? <filename>/libraryname?

Can you have two libraries/contracts of the same name within one Solidity file? i.e., 

```
import "SomeLib.sol";
import "./must/go/deeper/SomeLib.sol"; 
```

Assuming both define a Library called `SomeLib`, I assume the second import will overwrite first? If so, this means it's not the linker's problem - there's only one library to link. 

[...]

PR #8, opened 2016-05-11
Import basic bytecode linking code from browser-solidity

axic:
This is only the internal part of the linker, which replaces library references with addresses in a given bytecode. Code taken form browser-solidity. First step to implement #4.

The higher level method will need to support JSON input and perhaps handle multiple versions of the same library name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,3 +104,15 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
 	var output = solcSnapshot.compile("contract t { function g() {} }", 1);
 });
 ```
+
+### Linking bytecode
+
+When using libraries, the resulting bytecode will contain placeholders for the real addresses of the referenced libraries. These have to be updated, via a process called linking, before deploying the contract.
+
+The `linkBytecode` method provides a simple helper for linking:
+
+```javascript
+bytecode = solc.linkBytecode(bytecode, { 'MyLibrary': '0x123456...' });
+```
+
+Note: in future versions of Solidity a more sophisticated linker architecture will be introduced.  Once that changes, this method will still be usable for output created by old versions of Solidity.
--- a/test/package.js
+++ b/test/package.js
@@ -51,3 +51,50 @@ tape('Loading Legacy Versions', function (t) {
     });
   });
 });
+
+tape('Linking', function (t) {
+  t.test('link properly', function (st) {
+    var input = {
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -45,11 +45,25 @@ function setupMethods (soljson){
 		return JSON.parse(result);
 	}
 
+        var link = function(bytecode, libraries) {
+                for (var libraryName in libraries) {
+                        var libLabel = '__' + libraryName + Array(39 - libraryName.length).join('_');
```
chriseth: This should catch the case where `libraryName.length` is larger than 39.

index.js
```diff
@@ -45,11 +45,25 @@ function setupMethods (soljson){
...
+
+		return bytecode;
+        };
+
 	var version = soljson.cwrap("version", "string", []);
 
 	return {
 		version: version,
 		compile: compile,
+		link: link,
```
chriseth: I would prefer to find a different name for this to show that is a kind of "raw" way of linking and operating on LinkerObjects should be preferred.

axic: I don't have any good candidate. `linkBytecode`? `

index.js
axic: Actually what should happen in that case? What is happening within Solidity? Error or truncate?

axic: @chriseth ping

index.js
axic: @chriseth ping

index.js
axic: Well, just tried it out:

```js
library NameTooLong12345678XX12345678XX12345678XX12345678XX {
    function a() returns (bool) {
        return true;
    }
}

contract A {
    function a() returns (bool) {
        return NameTooLong12345678XX12345678XX12345678XX12345678XX.a();
    }
}
```

The output is: `__NameTooLong12345678XX12345678XX12345__`

It truncates it.

axic:
@chriseth this rebased and the library name is handled properly even when truncated.

Didn't come up with a better name yet, maybe `linkBytecode`?

axic:
@chriseth do you want to merge this so we have a solution until the `LinkerObject` is exposed? (And perhaps a solution for binaries generated with old soljson.)

chriseth:
Can you add documentation and unit tests?

axic:
@chriseth added

Issue #10, opened 2016-05-11
Latest 4 solc versions are missed.


Latrasis:
When listing the remote versions the loop misses the last 4 versions in the list.

```js
for (var i = sources.length - 1; i >= 0; i--) {
    // FIXME: use build as well
    var version = sources[i].match(/^soljson-v([0-9.]*)-.*.js$/);

    // Skip invalid lines
    if (!version) {
      continue;
    }

    if (version[1] === wanted) {
      downloadBinary(sources[i].match(/^soljson-(.*).js$/)[1]);
      return;
    }
  }
```

axic:
@Latrasis as an improvement it would be nice to improve solc-bin by having the versions in a JSON file. 

solc-bin contains [list.txt](https://github.com/ethereum/solc-bin/blob/gh-pages/bin/list.txt) and [list.js](https://github.com/ethereum/solc-bin/blob/gh-pages/bin/list.js) for the versions. The latter has two fields: `versions` (equals to list.txt) and `releases` (which is the first build of each version).

Do you think you could improve the process in solc-bin? That could help solc-js :)

Latrasis:
@axic: Sure thing! Just to understand the process, what is the process of adding a new version? Is the `list.txt` appended to? Or is it generated?

axic:
@Latrasis both are generated by [this script](https://github.com/ethereum/solc-bin/blob/gh-pages/update-index.sh)

chriseth:
Can you please explain a bit more why you think this is a bug?

Latrasis:
The very reason I stumbled upon this was when the latest version`3.2` got an update on `soljson-v0.3.2-2016-05-06-9e36bdd.js` which allowed inlined library contracts: https://github.com/ethereum/solidity/issues/493. This technically should have been `3.3`, but it was still `3.2`.

The realtime solidity compiler (RSC) does fetch latest version, however `solc-js` didn't which meant that any contract that would compile on RSC, but would not compile on `truffle` or other tools. Logically to me both should have the same latest version.

chriseth:
Ok, thanks, now I understand. The "inline library functions" feature is not yet released. It is part of the nightly developer snapshots that are used by browser-solidity. Version 0.3.3 will be released as soon as solidity is stable enough and it will of course contain this new feature.

The npm version is created for releases only in order to prevent people from deploying contracts with non-release versions (which are less stable and harder to verify because they might not be as well archived as the released versions).

[...]

PR #11, opened 2016-05-12
Fix Fetching Latest Remote Version

Latrasis:
See issue: #10


axic:
I think the simple solution is to iterate the list from top-to-bottom and not bottom-to-top.  See #12.

PR #12, opened 2016-05-12
Iterate the version list (list.txt) from top

### LINKED ISSUE #10: Latest 4 solc versions are missed.

axic:
Fixes #10

Or do we consider the first version as the main release and the subsequent ones as builds? If yes which one should be downloaded?

### DIFF
--- a/downloadCurrentVersion.js
+++ b/downloadCurrentVersion.js
@@ -40,7 +40,7 @@ getVersionList(function(list) {
   var wanted = package.version.match(/^(\d\.\d\.\d)-?\d?$/)[1];
 
   var sources = list.split('\n');
-  for (var i = sources.length - 1; i >= 0; i--) {
+  for (var i = 0; i < sources.length; i++) {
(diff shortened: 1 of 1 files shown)


chriseth:
I think the existing code is correct: The list contains all nighlty snapshots (or builds manually triggered for a release) and the "released" version is the first one (in the time sense).

chriseth:
Please reopen if you think your change is correct.

axic:
It all makes sense. Needs to be documented though.

PR #19, opened 2016-06-03
Add minimal testing.

chriseth:


axic:
For reference, this should add tests for the issue https://github.com/ethereum/solidity/issues/17 (actual fix: https://github.com/ethereum/solidity/pull/536)

PR #21, opened 2016-06-04
Split out setupMethod to wrapper.js

axic:
This way the wrapper can be included and browserified in projects without including the builtin soljson:

```
var solc = require('solc/wrapper');
// assuming soljson was loaded in the global context
var compiler = solc(Module);
```

(Is needed to accomplish the end goal of https://github.com/ethereum/browser-solidity/pull/54)


axic:
@chriseth rebased

Issue #24, opened 2016-06-06
Include an option or alternative method which passes the raw strings


axic:
and doesn't do JSON (de)serialisation.

It could be utilised by anyone using a browser worker, where only primitive types (strings) can be transmitted reliably. Without this option, that results in JSON encoding/decoding twice.

chriseth:
Can you explain a bit more which passing you are referring to?

axic:
```js
  var compile = function (input, optimise, readCallback) {
    var result = '';
    if (readCallback !== undefined && compileJSONCallback !== null) {
      result = compileJSONCallback(JSON.stringify(input), optimise, readCallback);
    } else if (typeof input !== 'string' && compileJSONMulti !== null) {
      result = compileJSONMulti(JSON.stringify(input), optimise);
    } else {
      result = compileJSON(input, optimise);
    }
    return JSON.parse(result);
  }
```

The compile method with JSON stringify & parse.  If solc is used within a web worker, the web worker will have to JSON stringify & parse again.

It would be nice to somehow have a feature where we can turn on/off the JSON stringify & parse.

chriseth:
Yes! My plan is actually to encode everything into json (string) until the very internals of the compiler - because that allows us to recompile easily for verification.

axic:
See comment: https://github.com/ethereum/solc-js/issues/33#issuecomment-314225490

PR #25, opened 2016-06-09
Export raw compile* methods too

axic:
Possible (partial) solution to #24

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -75,6 +75,9 @@ function setupMethods (soljson) {
   return {
     version: version,
     compile: compile,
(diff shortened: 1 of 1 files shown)


chriseth:
I think we should talk about the general architecture here, first. This seems like it clutters the module needlessely. At best, the exterior does not have to worry about which interface the compiler provides.

chriseth:
@axic by the way, this is the current proposal for the compiler input: https://github.com/ethereum/solidity/pull/645/files

axic:
I'm closing this for now.

Issue #9, opened 2016-05-11
Why doesn't this package include a cli?


raineorshine:
I was expecting I could `npm install -g solc` and have `solc` in my path. That was the impression I got from https://solidity.readthedocs.io/en/latest/installing-solidity.html:

> ## NPM / node.js
> This is probably the most portable and most convenient way to install Solidity locally.
> 
> A platform-independent JavaScript library is provided by compiling the C++ source into JavaScript using > Emscripten for browser-solidity and there is also an NPM package available.
> 
> To install it, simply use
> 
>     npm install solc

I guess I just install the [binaries](https://github.com/ethereum/webthree-umbrella/releases) instead?

raineorshine:
Or perhaps I install it with [these instructions](https://github.com/ethereum/go-ethereum/wiki/Contract-Tutorial) (brew on Mac):

```
brew install cpp-ethereum
brew linkapps cpp-ethereum
which solc
```

axic:
I think the documentation is misleading.  In theory a cli could be written on top of solc-js, but it would be limited in options compared to the C++ solc cli.

raineorshine:
Okay, thanks. Was a bit confusing.

Installing cpp-ethereum with brew worked.

redsquirrel:
@raineorshine I was confused about this too when I got started. Seems like there's an opportunity to clarify things on the npm/solc page.

chriseth:
Someone should extend the package :-)

InsidersByte:
@raineorshine @redsquirrel @chriseth @axic I created a simple cli https://github.com/InsidersByte/solc-cli

raineorshine:
@InsidersByte Great work!

I do think it may be less confusing for people if cli functionality was built into the `solc` package.

redsquirrel:
See #27

InsidersByte:
ha, I didn't realise there was an open pull request when I wrote it. Looks like I wrote it the day after it was created.

redsquirrel:
@InsidersByte Great minds think alike... at the same time! 😄

PR #27, opened 2016-06-09
Introduce cli - solcjs

axic:
Moving the compiler part of my old tool https://github.com/axic/ethereumjs-codesim.

Would it make sense mirroring the CLI of `solc` or should we diverge?

Note: we cannot 100% mirror as many features are lacking (as of today) in soljson.

It is very basic at the moment, only supports `--bin`. Most of the features (AST, ABI, interface) can be supported.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 var fs = require('fs');
+var path = require('path');
 var solc = require('./index.js');
 // FIXME: remove annoying exception catcher of Emscripten
 //        see https://github.com/chriseth/browser-solidity/issues/167
@@ -20,6 +21,11 @@ var yargs = require('yargs')
     describe: 'Binary of the contracts in hex.',
     type: 'boolean'
   })
+  .option('destination', {
+    alias: 'D',
+    describe: 'Output directory for the contracts.',
+    type: 'string'
+  })
   .global([ 'version', 'optimize' ])
   .version()
   .showHelpOnFail(false, 'Specify --help for available options')
@@ -28,6 +34,7 @@ var yargs = require('yargs')
 
 var argv = yargs.argv;
 var files = argv._;
+var destination = argv.destination || '.'
 
 function abort (msg) {
   console.log(msg || 'Error occured');
@@ -47,5 +54,5 @@ for (var i = 0; i < files.length; i++) {
 var output = solc.compile({ sources: sources }, argv.optimize ? 1 : 0);
 
 for (var contractName in output.contracts) {
-  fs.writeFileSync(contractName + '.bin', output.contracts[contractName].bytecode);
+  fs.writeFileSync(path.join(destination, contractName + '.bin'), output.contracts[contractName].bytecode);
 }
--- a/solcjs
+++ b/solcjs
@@ -21,8 +21,8 @@ var yargs = require('yargs')
     describe: 'Binary of the contracts in hex.',
     type: 'boolean'
   })
-  .option('destination', {
-    alias: 'D',
+  .option('output-dir', {
+    alias: 'o',
     describe: 'Output directory for the contracts.',
     type: 'string'
   })
@@ -34,7 +34,7 @@ var yargs = require('yargs')
 
 var argv = yargs.argv;
 var files = argv._;
-var destination = argv.destination || '.'
+var destination = argv['output-dir'] || '.'
 
 function abort (msg) {
   console.log(msg || 'Error occured');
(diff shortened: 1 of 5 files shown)


chriseth:
We probably do not want to move all the features over without rethinking the design first, but this is good to have, I guess.

solcjs
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
+
+var fs = require('fs');
+var solc = require('./index.js');
+// FIXME: remove annoying exception catcher of Emscripten
+//        see https://github.com/chriseth/browser-solidity/issues/167
+process.removeAllListeners('uncaughtException');
```
chriseth: Can we move that into index.js?

chriseth:
Please also add a test for the cli.

solcjs
raineorshine: Better to keep the shebang in a separate file. Then index.js can still be required when used programmatically.

chriseth:
oh, by the way: https://github.com/InsidersByte/solc-cli/blob/master/index.js

InsidersByte:
as I said in #9 I didn't realise this was open, when I wrote that cli. :smile:

InsidersByte:
one thing that would be useful (it's in my version) would be the ability to specify an output directory for the compiled contracts

solcjs
axic: I think @chriseth only meant the `removeAllListeners` part. That must only be called under node.js. For some reason it doesn't seem to work reliably. It is annoying bug discussed in https://github.com/chriseth/browser-solidity/issues/167, any help welcome solving it.

axic:
I think we should make a decision first whether the CLI should follow the C++ solc or be entirely different.

axic:
And decide on the name. Install it with `npm install -g solc` and you'll get it in the global path. It is currently `solcjs`. I think `solc` is a bad idea as it would collide with the C++ package.

axic:
(Just for bookkeeping: fixes #9)

solcjs
```diff
@@ -0,0 +1,51 @@
...
+var sources = {};
+
+for (var i = 0; i < files.length; i++) {
+  sources[ files[i] ] = fs.readFileSync(files[i]).toString();
+}
+
+var output = solc.compile({ sources: sources }, argv.optimize ? 1 : 0);
+
+for (var contractName in output.contracts) {
+  fs.writeFileSync(contractName + '.bin', output.contracts[contractName].bytecode);
```
chriseth: I'm not so sure it is a good idea to just write to arbitrary files - a clean output directory would probably be better.

axic: Yep, this is copying what `solc` does.

axic:
@chriseth I think this could be merged. It follows the parameter layout of `solc`.

@InsidersByte would you be interested in contributing more options to this later?

axic:
@Denton-L would you mind wrapping up some documentation for this?

Basic usage is as follows:

`solcjs -o output --bin --abi --interface MyToken.sol`

Current help screen:
```
Usage: solcjs [options] [input_file...]

Options:
  --version         Show version number                                [boolean]
  --optimize        Enable bytecode optimizer.                         [boolean]
  --bin             Binary of the contracts in hex.                    [boolean]
  --abi             ABI of the contracts.                              [boolean]
  --interface       Solidity Interface of the contrcats.               [boolean]
  --output-dir, -o  Output directory for the contracts.                 [string]
  --help            Show help                                          [boolean]
```

Denton-L:
Where do you want me to put the documentation?

axic:
@Denton-L you could do a separate PR or point me to your repo and I cherry pick. Thank you, it is much appreciated!

Denton-L:
I meant, where in the repository should I put the documentation?

axic:
@Denton-L in README.md

I've actually started the separation like this:

```
## Node.js Usage

<current text about npm install>

### Using on the command line

<command explanation>

### Using in projects

<current text here>
```

InsidersByte:
@axic IMO i would use the latest es6 / es7 features and transpile to es5 using [babel](https://babeljs.io/). 

That's what I did in [my cli](https://github.com/InsidersByte/solc-cli) if you want an example.

axic:
@InsidersByte thanks, I think we could do that.  I'll try to get this basic version merged and then we can improve it in further PRs.

chriseth:
We should allow remappings as a next feature.

axic:
@chriseth can you add that as an issue?

@InsidersByte it was merged, I'm happy to review PRs changing it.

@Denton-L I've added a small change to the readme, please create a PR if you would like to improve it.

Denton-L:
@axic, I don't think there's much to add, in terms of documentation, as `solcjs --help` covers everything pretty well.

PR #37, opened 2016-08-16
Version 0.3.6

chriseth:


axic:
Assuming the interface code is removed from Solidity in 0.4.0, perhaps we should remove `--interface` from the CLI before this release (as it was never released before).

chriseth:
Yes, it will be removed.

PR #45, opened 2016-09-05
Catch & rethrow exception so the function pointer can be freed

axic:
Aims to fix https://github.com/ethereum/browser-solidity/issues/212

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -16,7 +16,7 @@ function setupMethods (soljson) {
       soljson.setValue(ptr, buffer, '*');
     };
     var wrapCallback = function (callback) {
-      return soljson.Runtime.addFunction(function (path, contents, error) {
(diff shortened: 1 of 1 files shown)


chriseth:
What about adding / removing the function only when a new compiler version is loaded / unloaded? We can assign the callback to a variable available to the wrapper shortly before we call compileInternal.

chriseth:
CI failures due to style problem

axic:
> CI failures due to style problem

Fixed.

axic:
> What about adding / removing the function only when a new compiler version is loaded / unloaded? 

The only drawback is the need of a `cleanup()` method which needs to be called by the user. Also somehow in the case of multithreading it can cause issues.

chriseth:
Checked it out and wanted to fix it in the same second :-)

PR #52, opened 2016-09-09
Adds loadVersionFromFilePath and loadRemoteVersion returns script

kobigurk:


wrapper.js
```diff
@@ -99,13 +99,18 @@ function setupMethods (soljson) {
         } else {
           response.pipe(mem);
           response.on('end', function () {
-            cb(null, setupMethods(requireFromString(mem.toString())));
+            var solcContent = mem.toString();
+            cb(null, setupMethods(requireFromString(solcContent)), solcContent);
```
axic: What is the reason for changing `cb()` to 3 parameters?

kobigurk: The idea is supporting more interesting use cases rather than just requiring the file, such as caching. This method contains the logic of fetching it, so thought it's an appropriate place. Do you think a separate method would be better?

wrapper.js
```diff
@@ -99,13 +99,18 @@ function setupMethods (soljson) {
...
           });
         }
       }).on('error', function (error) {
         cb(error);
       });
     },
+    // Loads the compiler from a given file path. 
+    loadVersionFromFilePath: function (filePath) {
```
axic: This should probably be in line with the above method (`loadRemoteVersion`) and called `loadLocalVersion`.

axic:
I'm not entirely convinced this is needed since it can achieved with the following code: `let solc = require('solc/wrapper')(require(<filename>)`

It might be useful to have a shorthand for the above though (to reduce user mistakes and multiple reimplementations of the same), so let's consider it.

r0qs:
Thanks for the PR, but I'll be closing it due to inactivity. If it is still relevant, please ask for reopening.

PR #48, opened 2016-09-09
Add support to download custom compiler and use it from solcjs cli

chevdor:
Backward compatibility is maintained. The change allows listing all the versions,
getting a specific version, getting all the releases, all the non-nighlty or
simply all of them. Downloaded version are stored into a .bin/ folder that
will be automatically created.

Issue #51, opened 2016-09-09
Create output dir if it does not exist


chevdor:
Currently:

```
 solcjs -o **IdontExist** --bin contract.sol
```

fails because **IdontExist** well... does not exist.

Using fs-extra makes it an easy fix:

```
var fs = require('fs-extra);
fs.ensureDirSync (outputDir);
```

chevdor:
See #48

Issue #22, opened 2016-06-05
Document 'errors'


graup:
By trial and error I found out that in case the compilation fails `compile()` returns an object with a `errors` attribute, which is a list of all compiler errors. This is really useful! Maybe it could be documented somewhere. I'm not sure here is the best place, because that already comes from the compiler I assume, but [here](https://www.npmjs.com/package/solc) was where I was looking for it.

chevdor:
I will publish a commit in a second about that one.
btw: I totally agree!

chevdor:
Houston, we have a problem!
I do not want to change the tests without discussing a bit further the reason for them to specifically state that the compilation *should not* produce any output:

See https://github.com/ethereum/solc-js/blob/master/test/cli.js#L24 for instance:
```
 t.test('--bin', function (st) {
    var spt = spawn(st, './solcjs --bin test/DAO/Token.sol');
    spt.stderr.empty();
    spt.stdout.empty();        <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
    spt.succeeds();
    spt.end();
  });
```

It especially says that the output should be empty. So my fix in 990e2f5 break those tests.

I would ask someone to confirm that this part of the test can and should be remove, then I am happy to do so.

chevdor:
See #61

PR #55, opened 2016-09-23
PR48: Add support to download custom compiler and use it from solcjs cli

chevdor:
See #48 This is the same, just put in a separate branch.


chevdor:
Should be merged before #56

axic:
@chevdor so we should start with splitting up this PR. Let me know if you would have time for a quick chat on Gitter

axic:
We could start with creating PRs for (one for each item below):
- fix for #22
- fix for #51

chevdor:
#22 isolated in PR #61

chevdor:
#51 isolated in #62

chevdor:
Rebased version of this PR without #22 and #51 in #63

PR #59, opened 2016-10-05
Wrapper should expose its helpers even if not loading soljson

yann300:
allow loading wrapper without any preloaded compiler

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -5,6 +5,7 @@ var https = require('https');
 var MemoryStream = require('memorystream');
 
 function setupMethods (soljson) {
+  if (soljson) {
   var compileJSON = soljson.cwrap('compileJSON', 'string', ['string', 'number']);
   var compileJSONMulti = null;
   if ('_compileJSONMulti' in soljson) {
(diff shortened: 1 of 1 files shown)


chriseth:
This looks to me more like "force loading without compiler"

axic:
@yann300 as mentioned I had a different way of solving this.

Can you explain your actual use case?

yann300:
the compiler is used by remix's tests:
I want to be able to call `loadRemoteVersion()` version without having to specify a preloaded solc.js.
now `solcWrapper()` will throw.

`
var solcWrapper = require('solc/wrapper')

solcWrapper().loadRemoteVersion('latest', function (err, compiler) {
    if (err) {
      t.fail(err)
    } else {
      test(t, compiler)
    }
  })`

axic:
@yann300 thanks, that matches one of my motivations. I'll look into this today.

chriseth:
@yann300 @axic is this still needed?

axic:
I don't think this is needed since wrapper can be loaded independently. @yann300 please let us know and we can reopen.

PR #61, opened 2016-10-23
Document 'errors'

chevdor:


solcjs
```diff
@@ -59,6 +62,7 @@ for (var i = 0; i < files.length; i++) {
   }
 }
 
+console.log("Compiling with version: " + solc.version());
```
chriseth: Please do not clutter stdout

test/cli.js
```diff
@@ -11,42 +11,42 @@ tape('CLI', function (t) {
...
 
   t.test('invalid file specified', function (st) {
     var spt = spawn(st, './solcjs --bin test/fileNotFound.sol');
-    spt.stdout.match(/^Error reading /);
+    spt.stderr.match(/^Error reading /);
     spt.end();
   });
 
-  t.test('--abi', function (st) {
+  t.test('--abi -o output', function (st) {
```
chriseth: Please revert (see comment above).

wrapper.js
```diff
@@ -50,6 +50,23 @@ function setupMethods (soljson) {
...
     }
+
+    var res = JSON.parse(result);
+    if (res.errors) {
+      // TODO: Remove once the following is clarified:
+      // https://github.com/ethereum/solc-js/issues/53
+      // only leave the console.error
+
+      if (res.errors[0].indexOf('Warning') > 0) {
+        console.log('Compiled with Warnings: ', res.errors);
```
chriseth: Please output that on stderr.

chriseth:
In general, I think anything that is not regular output of the compiler should go to stderr. The fact whether the compilation was successful or not should be reported via the process exit status. If there are no errors (but warnings), exit status should report "success" but the stderr stream should be nonempty.

chriseth:
@chevdor could you please provide a summary of the PR as the title (and commit message) instead of just mentioning the issue number?

solcjs
```diff
@@ -41,10 +41,13 @@ var files = argv._;
 var destination = argv['output-dir'] || '.'
 
 function abort (msg) {
-  console.log(msg || 'Error occured');
+  console.error(msg || 'Error occured');
   process.exit(1);
 }
 
+if (!files.length)
+  abort('You must provide at least one file to compile.');
```
axic: This will need to be wrapper into `{}` to satisfy the lint.

test/cli.js
```diff
@@ -11,42 +11,42 @@ tape('CLI', function (t) {
 
   t.test('no parameters', function (st) {
     var spt = spawn(st, './solcjs');
-    spt.stderr.match(/^Must provide a file/);
+    spt.stderr.match(/^Must provide a file.*/);
```
axic: The `.*` shouldn't be needed as there's no closing `$` to mean end-of-match.

wrapper.js
```diff
@@ -50,6 +50,23 @@ function setupMethods (soljson) {
...
       result = compileJSON(input, optimise);
     }
+
+    var res = JSON.parse(result);
+    if (res.errors) {
+      // TODO: Remove once the following is clarified:
+      // https://github.com/ethereum/solc-js/issues/53
+      // only leave the console.error
+
+      if (res.errors[0].indexOf('Warning') > 0) {
```
axic: I think `res.errors[0]` is wrong, it should check every single error for being a warning.

derhuerst: Use `Array.prototype.filter`.

wrapper.js
```diff
@@ -50,6 +50,23 @@ function setupMethods (soljson) {
...
+        console.log('Compiled with Warnings: ', res.errors);
+      } else {
+        console.error('Compiled with Errors: ', res.errors);
```
derhuerst: Use `Array.prototype.filter`.

derhuerst:
@chevdor afaict you don't set the exit code to non-zero.

wrapper.js
```diff
@@ -50,6 +50,25 @@ function setupMethods (soljson) {
...
+    var res = JSON.parse(result);
+    if (res.errors) {
+      // TODO: Remove once the following is clarified:
+      // https://github.com/ethereum/solc-js/issues/53
+      // only leave the console.error
+
+      if (res.errors.filter(function (value) {
+        value.indexOf('Warning') > 0;
+      })) {
+        console.error('Compiled with Warnings: ', res.errors);
```
chriseth: Wrapper is too low-level to be allowed to write to the console.

solcjs
```diff
@@ -41,10 +41,14 @@ var files = argv._;
 var destination = argv['output-dir'] || '.'
 
 function abort (msg) {
-  console.log(msg || 'Error occured');
+  console.error(msg || 'Error occured');
   process.exit(1);
 }
 
+if (!files.length) {
+  abort('You must provide at least one file to compile.');
```
axic: This has been fixed since.

solcjs
```diff
@@ -41,10 +41,14 @@ var files = argv._;
 var destination = argv['output-dir'] || '.'
 
 function abort (msg) {
-  console.log(msg || 'Error occured');
+  console.error(msg || 'Error occured');
```
axic: The error/log change is handled by #129.

axic:
Closing this as it seems everything has been addressed since.

PR #62, opened 2016-10-23
Always create output directory

### LINKED ISSUE #51: Create output dir if it does not exist

chevdor:
Fixes #51 
Create output folder if it does not exist

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   },
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
+    "fs-extra": "^0.30.0",
     "memorystream": "^0.3.1",
     "require-from-string": "^1.1.0",
     "yargs": "^4.7.1"
--- a/solcjs
+++ b/solcjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-var fs = require('fs');
+var fs = require('fs-extra');
 var path = require('path');
 var solc = require('./index.js');
 // FIXME: remove annoying exception catcher of Emscripten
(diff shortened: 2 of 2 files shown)


chriseth:
I'm not sure this is wanted behaviour.

chriseth:
Could you please explain the use-case of this change? I would say it is better to fail in this case.

chevdor:
Unlike solc, solcjs does not provide the compilation results on stdout but in a folder. This folder is mandatory.

In the other PRs, the default value for the ouput has been changed from being nothing (ie current folder, which creates a mess) to a directory called 'out', unless the user specifies another value.

So by default, unless the user creates the output folder ('out'), compilation would always fail without this PR.

chriseth:
Thanks for the explanation! Note that solc can also operate in the mode where an output folder is used and this is the only real way to use it in a context where machine readability is required.

I think it would be cleaner not to use any default output folder but rather fail if no output folder is specified or if the output folder does not exist. We should get this as close to solc as possible.

chevdor:
Hello @chriseth,

Thanks for your comments. The point is not that solc can output file, it is that solc-js **must** output files. So not creating the default directory totally makes no sense.

While, we don´t **have to** create the out dir, not doing so is a bit like when you get your new computer and you get an error right away telling you that you need to create your home directory before things can start working. That would not be a very pleasant experience.

That the reason why I decided to create the output folder **if it does not exist yet**.
In the end, this is no drawback for anyone. For the people who are used to create the directory, it will just work as before. For the others, who did not think about it at first, the directory will be created. Users who want more control can specify where they want the output as well.

With all respect, I humbly disagree with the change you request. I see no reason to fail when we can avoid the failure and allow users to get started with solc-js quicker and easier.
This also makes the tests much cleaner btw.

chriseth:
I'm not sure if everyone would agree with your statement that writing the output files to the current directory would create a mess. I think I'm "more fine" with force-creating the output directory than with changing the default output directory (that is a strong breaking change, actually) from "." to "out"

chevdor:
Any plan on making a decision on this PR?

PR #63, opened 2016-10-23
Add support for options

chevdor:
This is a rebased version of #55 

- Call with --help
- New option to supply a custom compiler from the CLI
- Add very basic contract
- Move code out of the wrapper
- Add clean script
- Add some doc


chriseth:
I think the option to use a specific version of the compiler is a nice feature. What do you think, @axic? I would suggest `--use-version <version>` where version is a semver version (without the `soljson` prefix).

chevdor:
@chriseth regarding your last comment, I initially kept the +soljson+ part initially to allow easy copy past from the version list. No strong opinion on that though. Could you please provide feedback and/or merge?

axic:
@chevdor it would still be nice to merge your changes but they must be split up into bite sized chunks.

If you are still interested, could you create a PR which *only contains* `solcjs-get`?

chevdor:
Hi Alex
It has been a while. I need to see what is in those PRs. I see think it
would be great to have.

tzapu:
hey guys, thank you for your wonderful work. are there anymore plans to include this? i would really really love to be able to specify compiler version on the cli

cheers

chevdor:
Thanks for the bump @tzapu 
I am checking out those (ancient) conflicts :)

chevdor:
@axic I made a PR with only solcjs-get. See https://github.com/ethereum/solc-js/pull/227

tzapu:
thanks @chevdor for picking this right up
one thought, should `loadRemoteVersion` not use a version that's already downloaded, if it exists?
it would allow you to used cached versions of the bin via  `solcjs-get` without having to download every time you run

chevdor:
I don´t mind improving but I would only invest the time if some of the PRs I submitted years ago get merged. 

I created #227 to allow starting with smaller chunks and I would rather not grow the current PR too much right now as the reason it was not merged in the first place is probably the size.

So I suggest focusing on #227 at first, once we have this one in, #63 will probably be easier to review as it mainly contains additional tests and doc.

chevdor:
Ancient PR, not willing to fix conflicts once again.

PR #64, opened 2016-10-23
Major refactoring

chevdor:
Add support for local list
Fix doc
Remove solcjs-get
New tests
Refactoring of the tests
Fix cli-build tests
Add new test contract for errors
Fix for building with a custom compiler
Fix to replace the local soljson.js

End result:
[asciicast](https://asciinema.org/a/86834)


chevdor:
We need to merge #63 before we can consider this PR.

.gitignore
```diff
@@ -25,3 +25,11 @@ build/Release
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+<<<<<<< HEAD
```
chriseth: Merge conflict

README.md
```diff
@@ -23,6 +23,22 @@ To see all the supported features, execute:
 solcjs --help
 ```
 
+<<<<<<< HEAD
```
chriseth: Merge conflict

package.json
```diff
@@ -33,7 +39,13 @@
   },
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
+<<<<<<< HEAD
```
chriseth: Merge conflict

chevdor:
@chriseth to build a file called get you don´t use `solcjs get` but something like `solcjs build get`.
Moreover, `--bin` or `--abi` is required so we end up with `solcjs build --abi get` for instance.

index.js
```diff
@@ -1,3 +1,9 @@
-var wrapper = require('./wrapper.js');
+var fs = require('fs');
 
-module.exports = wrapper(require('./soljson.js'));
+try {
+  fs.lstatSync('./soljson.js');
+  var wrapper = require('./wrapper.js');
+  module.exports = wrapper(require('./soljson.js'));
+} catch (e) {
+  console.error('soljson.js not available.');
```
chriseth: I think we should not assume that `index.js` is used from a context that has a console available (the same applies to the wrapper). The whole package may be used through browserify.

lib/compilers.js
```diff
@@ -0,0 +1,208 @@
...
+          var i = builds[item];
+
+          if (i.prerelease) {
+            n.push(i);
+          }
+        }
+        return n;
+      };
+
+      // not a valid way...
```
chriseth: Please remove.

lib/helpers.js
```diff
@@ -0,0 +1,83 @@
+var fs = require('fs-extra');
+var chalk = require('chalk');
+
+function helpers () {
+  function extractCompilerData (path) {
+    var res = {};
+    var match = /v\d\.\d\.\d/.exec(path);
+    if (match) { res.version = match[0]; }
+
+    match = /(nightly.+)\+/.exec(path);
```
chriseth: The regex should be something like `/-([^+]*)/`

lib/helpers.js
```diff
@@ -0,0 +1,83 @@
...
+
+    if (match) {
+      res.prerelease = match[1];
+    }
+
+    match = /(commit\.\w+)\./.exec(path);
```
chriseth: And here it should be `/+(.*)/`

solcjs
```diff
@@ -1,72 +1,327 @@
...
+  default: false,
+  describe: 'Build - Generate the bytecode of the contract(s)'
+})
+.option('abi', {
+  describe: 'Build - Generate the ABI of the contract(s)',
+  type: 'boolean'
+})
+.option('output-dir', {
+  alias: 'o',
+  describe: 'Build - Output directory. Defaults to ./out',
```
chriseth: Please change to `.`.

PR #93, opened 2017-04-21
Support compileStandard

axic:


wrapper.js
```diff
@@ -26,19 +27,32 @@ function setupMethods (soljson) {
         }
       };
     };
-    var compileInternal = soljson.cwrap('compileJSONCallback', 'string', ['string', 'number', 'number']);
-    compileJSONCallback = function (input, optimize, readCallback) {
+
+    var runWithReadCallback = function (readCallback, compile) {
```
chriseth: Please add a comment that this accepts optional parameters.

axic:
@chriseth with the binary you've compiled all of the tests fail with:
```
{
  "errors": [
    {
      "component": "general",
      "formattedMessage": "Internal exception in StandardCompiler::compileInternal",
      "message": "Internal exception in StandardCompiler::compileInternal",
      "severity": "error",
      "type": "InternalCompilerError"
    }
  ]
}
```

axic:
It is a StandardCompiler implementation shortcoming:
```
libc++abi.dylib: terminating with uncaught exception of type Json::LogicError: in Json::Value::find(key, end, found): requires objectValue or nullValue
```

chriseth:
Can you add `lineInfo()` to anything that catches `Exception` and provides a generic error message?

axic:
I have fixed it in solidity, just haven't pushed the PR yet. The above message is with `catch (...)` so not sure what can be printed there.

chriseth:
@axic ah right, it is an exception from json, so we are out of luck.

axic:
This works properly for me. The failure on travis seems to be that `semistandard` is now incompatible with node 0.10, so we need to disable 0.10.

axic:
Depends on #94.

Issue #33, opened 2016-08-01
Change compile() so that it always works with JSON input


axic:
I think going forward, `compile()` should always work JSON input and the readFile callback and should always return a JSON.

`compile()` then should do the wrapping to work with older versions of soljson, which lack `compileJSONCallback` and/or `compileJSONMulti`.

axic:
I think this has been superseded by `compileStandard` and `compilerStandardWrapper`. In a future breaking change we should revamp the API though.

PR #95, opened 2017-04-25
Implement JSON I/O wrapper for older compilers

axic:
Depends on #93. Deprecates #33.

Todo:
- [x] Support AST
- [x] Translate methodIdentifiers
- [x] Translate gasEstimates
- [x] Fix `filename:contractname` parsing
- [x] Parse warnings vs. errors

### DIFF
--- a/translate.js
+++ b/translate.js
@@ -13,7 +13,50 @@ function translateJsonCompiler (output) {
     });
   }
 
-  // FIXME: translate actual output too
+  ret['contracts'] = {};
+  for (var contract in output['contracts']) {
+    // Split name first, can be `contract`, `:contract` or `filename:contract`
+    var tmp = contract.match(/^([^:]*):?([^:]+)$/);
+    var fileName, contractName;
+    if (tmp.length === 3) {
+      fileName = tmp[1];
+      contractName = tmp[2];
+    } else if (tmp.length === 2) {
+      fileName = '';
+      contractName = tmp[1];
+    } else {
+      // Force abort
+      return null;
+    }
+
+    var contractInput = output['contracts'][contract];
+    var contractOutput = {
+      'abi': contractInput['interface'],
+      'metadata': contractInput['metadata'],
+      'evm': {
+        'legacyAssembly': contractInput['assembly'],
+        'bytecode': {
+          'object': contractInput['bytecode'],
+          'opcodes': contractInput['opcodes'],
+          'sourceMap': contractInput['srcmap']
+        },
+        'deployedBytecode': {
+          'object': contractInput['runtimeBytecode'],
+          'sourceMap': contractInput['srcmapRuntime']
+        },
+        // FIXME
+        'methodIdentifiers': contractInput['methodIdentifiers'],
+        // FIXME translate
+        'gasEstimates': contractInput['gasEstimates']
+      }
+    };
+
+    if (!ret['contracts'][fileName]) {
+      ret['contracts'][fileName] = {};
+    }
+
+    ret['contracts'][fileName][contractName] = contractOutput;
+  }
 
   return ret;
 }
--- a/translate.js
+++ b/translate.js
@@ -21,10 +21,17 @@ function translateErrors (ret, errors) {
 }
 
 function translateGasEstimates (gasEstimates) {
+  if (gasEstimates === null) {
+    return 'infinite'
+  }
+
+  if (typeof gasEstimates === 'number') {
+    return gasEstimates.toString()
+  }
+
   var gasEstimatesTranslated = {}
   for (var func in gasEstimates) {
-    var estimate = gasEstimates[func]
-    gasEstimatesTranslated[func] = estimate !== null ? estimate.toString() : 'infinite'
+    gasEstimatesTranslated[func] = translateGasEstimates(gasEstimates[func])
   }
   return gasEstimatesTranslated
 }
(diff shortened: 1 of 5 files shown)


translate.js
```diff
@@ -0,0 +1,70 @@
...
+      formattedMessage: output['errors'][error]
+    });
+  }
+
+  ret['contracts'] = {}
+  for (var contract in output['contracts']) {
+    // Split name first, can be `contract`, `:contract` or `filename:contract`
+    var tmp = contract.match(/^([^:]*):?([^:]+)$/)
+    var fileName, contractName
+    if (tmp.length == 3) {
```
chriseth: This is always 3 for me. What about using `^(([^:]*):)?([^:]+)$/` - then filename is always `match[2]` and contract name is always `match[3]`

axic: `^([^:]*:)?([^:]+)$/` I guess. There's no need for two captures inside?

translate.js
```diff
@@ -0,0 +1,71 @@
...
+  if (output['formal']) {
+    ret['why3'] = output['formal']['why3'];
+    // FIXME: map errors to the above list
+  }
+
+  return ret;
+}
+
+module.exports = {
+  translateJsonCompiler: translateJsonCompiler
```
chriseth: This only translates the output, not the input, right? If yes, can we change the name accordingly?

translate.js
chriseth: There is, otherwise the filename is suffixed by `:`.

axic:
Need to add support for AST too.

wrapper.js
```diff
@@ -67,6 +68,81 @@ function setupMethods (soljson) {
...
+    if (input['language'] !== 'Solidity') {
+      return formatFatalError('Only Solidity sources are supported');
+    }
+
+    if (input['sources'] == null) {
+      return formatFatalError('No input specified');
+    }
+
+    // Bail out early
+    if ((input['sources'].length > 0) && (compileJSONMulti === null)) {
```
axic: Should be `length > 1`.

axic:
Need to include `translate.js` in package.json.

translate.js
axic: So the match results will be:
- first match is filename without `:`
- second match is filename with `:`
- third match if contractname?

translate.js
```diff
@@ -47,10 +47,7 @@ function translateJsonCompilerOutput (output) {
           'object': contractInput['runtimeBytecode'],
           'sourceMap': contractInput['srcmapRuntime']
         },
-        // FIXME
-        'methodIdentifiers': contractInput['methodIdentifiers'],
-        // FIXME translate
-        'gasEstimates': contractInput['gasEstimates']
+        'methodIdentifiers': contractInput['functionHashes'],
```
axic: I think the key/value order differs between the old JSON and the new JSON?

translate.js
```diff
@@ -48,6 +49,14 @@ function translateJsonCompilerOutput (output) {
           'sourceMap': contractInput['srcmapRuntime']
         },
         'methodIdentifiers': contractInput['functionHashes'],
+        'gasEstimates': {
+          'creation': {
+            'dataCost': gasEstimates['creation'][1],
+            'executionCost': gasEstimates['creation'][0]
+          },
+          'internal': gasEstimates['internal'],
+          'external': gasEstimates['external']
```
axic: The values are different between the old and the new JSON.

The new JSON has the gas values as string and also can contain `"infinite"`.

The old JSON uses `null` for infinite and anything which exceeds a Javascript number (>53 bits).

translate.js
```diff
@@ -72,6 +72,20 @@ function translateJsonCompilerOutput (output) {
...
+  var sourceMap = {}
+  for (var sourceId in output['sourceList']) {
+    sourceMap[output['sourceList'][sourceId]] = sourceId;
+  }
+
+  ret['sources'] = {}
+  for (var source in output['sources']) {
+    ret['sources'][source] = {
+      id: sourceMap[source],
+      ast: output['sources'][source],
```
axic: Please remove this, the old JSON cannot support the "new AST".

translate.js
yann300: what do you mean by order?
here is the old way
```
functionHashes
get(uint256,uint256):"669e48aa"
set(uint256):"60fe47b1"
```

translate.js
yann300: as string? that's not what is in http://solidity.readthedocs.io/en/develop/using-the-compiler.html#compiler-input-and-output-json-description

axic: Good catch, need to update that :)

translate.js
```diff
@@ -1,10 +1,19 @@
 function translateErrors (ret, errors) {
   for (var error in errors) {
-    // FIXME: parse warnings here
+    var type = 'error'
+    var extractType = /^(.*):(\d+):(\d+):(.*):/
```
axic: Is this the code from browser-solidity? I realised we had a code for it there.

translate.js
axic: Fixed in https://github.com/ethereum/solidity/pull/2188

translate.js
```diff
@@ -0,0 +1,113 @@
...
+          'sourceMap': contractInput['srcmap']
+        },
+        'deployedBytecode': {
+          'object': contractInput['runtimeBytecode'],
+          'sourceMap': contractInput['srcmapRuntime']
+        },
+        'methodIdentifiers': contractInput['functionHashes'],
+        'gasEstimates': {
+          'creation': {
+            'dataCost': gasEstimates['creation'][1] !== null ? gasEstimates['creation'][1].toString() : 'infinite',
```
axic: Actually this was renamed `codeDepositCost`.

axic:
I did a manual comparison now and it seems to be ok. Though we should have the tests run with all solc releases at some point :)

translate.js
```diff
@@ -0,0 +1,115 @@
...
+      type = extractType[4].trim().toLowerCase()
+    } else if (errors[error].toLowerCase().indexOf(': warning:')) {
+      type = 'warning'
+    } else if (errors[error].toLowerCase().indexOf(': error:')) {
+      type = 'error'
+    }
+    ret.push({
+      type: 'Error',
+      component: 'general',
+      severity: type,
```
axic: @yann300 `severity` must be either `'warning'` or `'error'`, while `type` should be the readable type.

translate.js
```diff
@@ -4,16 +4,16 @@ function translateErrors (ret, errors) {
...
+      type = 'Warning'
+    } else if (errors[error].indexOf(': Error:')) {
+      type = 'Error'
     }
     ret.push({
-      type: 'Error',
+      type: type,
       component: 'general',
-      severity: type,
+      severity: type.toLowerCase(),
```
axic: I think this should be `(type === 'Warning') ? 'warning' : 'error'`

PR #99, opened 2017-04-25
Support JSON I/O in CLI

### LINKED ISSUE #67: coherent output between solc and solcjs
Is there a reason why which are output by the native solc compiler differ from the solcjs output?
It would be great to receive the same output from both.

e.g. 
bytecode vs. bin
runtimeBytecode vs. bin-runtime
interface vs. abi

axic:
Fixes #67 and #97.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -28,6 +28,10 @@ var yargs = require('yargs')
     describe: 'ABI of the contracts.',
     type: 'boolean'
   })
+  .option('standard-json', {
+    describe: 'Turn on Standard JSON Input / Output mode.',
+    type: 'boolean'
+  })
   .option('output-dir', {
     alias: 'o',
     describe: 'Output directory for the contracts.',
@@ -37,7 +41,6 @@ var yargs = require('yargs')
   .version(function() { return solc.version(); })
   .showHelpOnFail(false, 'Specify --help for available options')
   .help()
-  .demand(1, 'Must provide a file');
 
 var argv = yargs.argv;
 var files = argv._;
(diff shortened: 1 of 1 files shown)


solcjs
```diff
@@ -45,6 +48,26 @@ function abort (msg) {
   process.exit(1);
 }
 
+if (argv['standard-json']) {
+  if (!solc.supportsStandard) {
+    abort('Compiler does not support Standard JSON I/O');
+  }
+
+  var size = fs.fstatSync(process.stdin.fd).size;
```
chriseth: Are you able to do that on a stream? Will it store the data somewhere?

solcjs
```diff
@@ -45,6 +48,26 @@ function abort (msg) {
...
+
+  if (size <= 0) {
+    abort('Empty input was read');
+  }
+
+  var input = fs.readSync(process.stdin.fd, size)[0];
+
+  console.log(solc.compileStandard(input));
```
chriseth: Shouldn't we use `process.stdout` here?

axic: `console.log` goes to stdout, `console.error` is stderr

solcjs
axic: Don't know, took this code from stackoverflow and it works (it counts till EOF).

chriseth: I don't think we should use `process.stdin.fd`: https://github.com/nodejs/node/issues/7439#issuecomment-228760636

solcjs
chriseth: ok!

solcjs
chriseth: Ok, so let's keep that and hope that it works for everyone :-)

PR #103, opened 2017-04-26
Verify version matches in prepublish

### LINKED ISSUE #1: Versioning
Currently the versioning of `solc` is not 100% [semver](http://semver.org/) (and thus npm) compatible.

e.g. `0.3.1-1` translates to _prerelease version_ of 1.

For reference, the C solc has a version of `v0.3.2-2016-04-22-dd4300d`, which includes a lot of build metadata. The right way to include that is by using the plus sign: `0.3.1+1` or `0.3.1+2016-04-22-dd4300d`.

@chriseth also raised that mirroring the version in JS causes the issue that interface changes in JS cannot be properly signaled using semver as it is tied to the C version.

axic:
Fixes #1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "semistandard",
-    "prepublish": "./downloadCurrentVersion.js",
+    "prepublish": "./downloadCurrentVersion.js && ./verifyVersion.js",
     "test": "tape ./test/index.js"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


axic:
```
$ ./verifyVersion.js 
solcVersion: 0.4.10+commit.f0d539ae.Emscripten.clang
packageVersion: 0.4.10
Version matching
$ echo $?
0
$ cp soljson.js.chris soljson.js 
$ ./verifyVersion.js 
solcVersion: 0.4.11-develop.2017.4.24+commit.9e8cc01b.Emscripten.clang
packageVersion: 0.4.10
Version mismatch
$ echo $?
1
```

PR #118, opened 2017-07-18
Add two tests for invalid Solidity input

axic:


test/package.js
```diff
@@ -12,9 +12,15 @@ tape('Compilation', function (t) {
...
     var output = solc.compile('contract x { this is an invalid contract }');
+    st.plan(3);
     st.ok('errors' in output);
-    st.equal(output.errors.length, 1);
-    st.notEqual(output.errors[0].indexOf('ParserError'), -1);
+    // Check if the ParserError exists, but allow others too
+    st.ok(output.errors.length >= 1);
+    for (var error in output.errors) {
+      if (output.errors[error].indexOf('ParserError') !== -1) {
+        st.ok(true);
```
yann300: but we want to fail if the ParserError is not present right?

axic: Yes. The `st.plan(3)` enforces there are three`.ok` calls, otherwise it fails.

Tested it by removing this for loop and it fails on "plan count mismatch".

yann300: ah ok

Issue #120, opened 2017-07-18
Rework the API in a breaking release


axic:
Currently there is:
- `compile` which wraps `compileJSON`, `compileJSONMulti` and `compileJSONCallback`
- `compileStandard` which wraps `compileStandard`
- `compileStandardWrapper` which uses or emulates `compileStandard`

I think the only method should be:
- `compile` which provide the standard JSON interface (uses `compileStandard` or emulates it`)

axic:
What is left: change `solcjs` to always use Standard JSON.

Tectract:
This caused a breaking change in browser-solc, and I'm trying to figure out how to fix it...

https://github.com/ericxtang/browser-solc/issues/4

RuntimeError: abort(Assertion failed: Cannot call unknown function compileJSON, make sure it is exported). Build with -s ASSERTIONS=1 for more info.

argh, compiler.compiler no longer works

PR #122, opened 2017-07-18
[WIP] Support the standard JSON via compile() too

axic:
Part of #120 and https://github.com/ethereum/solidity/issues/3452

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,19 @@ There is also a direct method, `compileStandard`, which is only present on recen
 
 Starting from version 0.4.20 a Semver compatible version number can be retrieved on every compiler release, including old ones, using the `semver()` method.
 
+#### From version 0.4.21
+
+Starting from version 0.4.21 a *standard JSON input and output* is also accepted by `compile()`:
+
+```javascript
+function findImports(path) {
+	/* Implement support for loading external files here */
+	return { error: 'File not found' }
+}
+var output = solc.compile({ /* standard JSON input */ }, findImports)
+/* output contains standard JSON output */
+```
+
 ### Using with Electron
 
 **Note:**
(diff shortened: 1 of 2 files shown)


axic:
Needs tests.

wrapper.js
```diff
@@ -68,6 +68,28 @@ function setupMethods (soljson) {
...
+        input = JSON.parse(input);
+        if (typeof input['language'] === 'string') {
+          isStandardJSON = true;
+        }
+      } catch (e) {
+      }
+
+      if (isStandardJSON) {
+        // NOTE: takes second argument as "readCallback"
+        return JSON.parse(compileStandardWrapper(input, optimise));
```
chriseth: `compileStandardWrapper` is only assigned in line 83 / 105

axic:
> I am not convinced this is working.

Probably it isn't yet, it is more about discussing the API.

Eventually at 0.5.0 I'd like to replace `compile` to only do what `compileStandardWrapper` is doing. This meant to be a stepping stone to there.

chriseth:
Yes, I think it is a good idea to make compile to also take json-io.

wrapper.js
axic: Javascript scoping... 😉

chriseth: Ah, it's a callback!

axic:
This needs tests.

README.md
```diff
@@ -104,6 +104,19 @@ There is also a direct method, `compileStandard`, which is only present on recen
 
 Starting from version 0.4.20 a Semver compatible version number can be retrieved on every compiler release, including old ones, using the `semver()` method.
 
+#### From version 0.4.21
```
axic: Oh man, only if we'd have merged this at 0.4.21, it wouldn't be too big a breaking change anymore :)

PR #124, opened 2017-07-19
Replace ':' (and `.` `/`) with '-' in output filenames for solcjs

### LINKED ISSUE #109: solcjs doesn't appear to do anything on Windows
Either I'm doing something wrong (the usage instructions are unclear), or there's a bug:

I just installed latest solcjs `npm install -g solcjs` on my Windows 10 machine, trying to compile a simple script.

Running 

```solcjs --bin Coin.sol``` thinks for a bit, and doesn't actually do anything, where Coin.sol is the following file:

```
pragma solidity ^0.4.0;

contract Coin {
    // The keyword "public" makes those variables
    // readable from outside.
    address public minter;
    mapping (address => uint) public balances;

    // Events allow light clients to react on
    // changes efficiently.
    event Sent(address from, address to, uint amount);

    // This is the constructor whose code is
    // run only when the contract is created.
    function Coin() {
        minter = msg.sender;
    }

    function mint(address receiver, uint amount) {
        if (msg.sender != minter) return;
        balances[receiver] += amount;
    }

    function send(address receiver, uint amount) {
        if (balances[msg.sender] < amount) return;
        balances[msg.sender] -= amount;
        balances[receiver] += amount;
        Sent(msg.sender, receiver, amount);
    }
}
```

solcjs version is ```0.4.11+commit.68ef5810.Emscripten.clang```

### PR DESCRIPTION
**Pull request by barchiesi**
On Windows systems, the ':' (colon) character is reserved and invalid in file names. This patch adds the --replace-colon option that replaces the colon with '-' (minus) in output file names.

Should fix #109

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -37,8 +37,8 @@ var yargs = require('yargs')
     describe: 'Output directory for the contracts.',
     type: 'string'
   })
-  .option('replace-colon', {
-    describe: 'Replace colon char with minus in output file names,',
+  .option('friendly-filenames', {
+    describe: 'Replace illegal filename chars with underscore in output file names,',
     type: 'boolean'
   })
   .global([ 'version', 'optimize' ])
@@ -111,16 +111,16 @@ function writeFile (file, content) {
 
 for (var contractName in output.contracts) {
   var contractFileName = contractName;
-  if (argv.replaceColon) {
-      contractFileName = contractFileName.replace(':', '-');
+  if (argv.friendlyFilenames) {
+      contractFileName = contractFileName.replace(/[:./]/g,'');
   }
 
   if (argv.bin) {
-    writeFile(contractName + '.bin', output.contracts[contractName].bytecode);
+    writeFile(contractFileName + '.bin', output.contracts[contractName].bytecode);
   }
 
   if (argv.abi) {
-    writeFile(contractName + '.abi', output.contracts[contractName].interface);
+    writeFile(contractFileName + '.abi', output.contracts[contractName].interface);
   }
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by axic** [comment]
Can you please copy what `solc` is doing? (https://github.com/ethereum/solidity/blob/develop/libsolidity/interface/CompilerStack.cpp#L352-L371)

It replaces `/`, `:` and `.` in the `contractName` with `_`.

File: solcjs [hunk 1]
```diff
@@ -96,6 +100,11 @@ if (!output) {
 
 fs.ensureDirSync (destination);
 for (var contractName in output.contracts) {
+  var contractFileName = contractName;
+  if (argv.replaceColon) {
+      contractFileName = contractFileName.replace(':', '-');
+  }
+
   if (argv.bin) {
     fs.writeFileSync(path.join(destination, contractName + '.bin'), output.contracts[contractName].bytecode);
```
axic: Need to use `contractFileName` here and below.

axic:
Thanks, can you also make it non-optional? (The same as in `solc`.)

axic:
Also please rebase :)

solcjs
```diff
@@ -96,12 +100,17 @@ if (!output) {
 
 fs.ensureDirSync (destination);
 for (var contractName in output.contracts) {
+  var contractFileName = contractName;
+  if (argv.friendlyFilenames) {
+      contractFileName = contractFileName.replace(/[:./]/g,'');
```
barchiesi: Must of pressed undo for some reason...

axic:
Thank you!

PR #127, opened 2017-07-25
Fix --standard-json on solcjs

### LINKED ISSUE #126: solcjs --standard-json - TypeError: Second argument needs to be a buffer
Hello,

`solcjs --standard-json myinput.json` doesn't work. See the error bellow. 

```
$ echo "{}" | solcjs --standard-json
/usr/local/lib/node_modules/solc/node_modules/graceful-fs/polyfills.js:144
        throw er
        ^

TypeError: Second argument needs to be a buffer
    at Object.fs.readSync (fs.js:681:18)
    at Object.readSync (/usr/local/lib/node_modules/solc/node_modules/graceful-fs/polyfills.js:138:28)
    at Object.<anonymous> (/usr/local/lib/node_modules/solc/solcjs:65:18)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
```

axic:
Fixes #126.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -51,15 +51,8 @@ function abort (msg) {
 }
 
 if (argv['standard-json']) {
-  var size = fs.fstatSync(process.stdin.fd).size;
-
-  if (size <= 0) {
-    abort('Empty input was read');
-  }
-
-  var input = fs.readSync(process.stdin.fd, size)[0];
-
-  console.log(solc.compileStandardWrapper(input));
+  var input = fs.readFileSync(process.stdin.fd);
+  console.log(solc.compile(input.toString('utf8')))
   process.exit(0);
 } else if (files.length === 0) {
   console.error('Must provide a file');
--- a/test/cli.js
+++ b/test/cli.js
@@ -62,4 +62,33 @@ tape('CLI', function (t) {
     spt.succeeds();
     spt.end();
   });
+
+  t.test('standard json', function (st) {
+    var input = {
+      'language': 'Solidity',
+      'settings': {
+        'outputSelection': {
+          '*': {
+            '*': [ 'evm.bytecode', 'userdoc' ]
+          }
+        }
+      },
+      'sources': {
+        'Contract.sol': {
+          'content': 'pragma solidity ^0.5.0; contract Contract { function f() pure public {} }'
+        }
+      }
+    };
+    var spt = spawn(st, './solcjs --standard-json');
+    spt.stdin.setEncoding('utf-8');
+    spt.stdin.write(JSON.stringify(input));
+    spt.stdin.end();
+    spt.stdin.on('finish', function () {
+      spt.stderr.empty();
+      spt.stdout.match(/Contract.sol/);
+      spt.stdout.match(/userdoc/);
+      spt.succeeds();
+      spt.end();
+    });
+  });
 });
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,6 @@ matrix:
   fast_finish: true
   include:
     - os: linux
-      node_js: '4'
-      env: CXX=g++-4.8 TEST_SUITE=test
-    - os: linux
       node_js: '6'
       env: CXX=g++-4.8 TEST_SUITE=test
     - os: linux


chriseth:
Tests are failing.

axic:
Not sure yet why, it works locally on my computer.

axic:
Fixed in `spt.stdin.end()` was prematurely closing the pipe, before travis (or any other slower system) managed to push the actual json.

test/cli.js
```diff
@@ -47,4 +47,24 @@ tape('CLI', function (t) {
...
+    };
+    var spt = spawn(st, './solcjs --standard-json');
+    spt.stdin.setEncoding('utf-8');
+    spt.stdin.write(JSON.stringify(input), function () {
+      spt.stderr.empty();
+      spt.stdout.match(/Contract.sol/);
+      spt.stdout.match(/userdoc/);
+      spt.succeeds();
+      spt.end();
+    });
```
watilde: You might need `grep.stdin.end();` around here.

axic: This version uses the callback of write which is executed when the stream was pushed. If you look at the commit history, all combinations were tried (pause, end, etc.) and none of them work on Travis. All of them work for me locally.

Any help is appreciated, it is really frustrating at this point.

watilde: Oh yeah you've already tried at https://github.com/ethereum/solc-js/pull/127/commits/5bef54c21570c412c39e00b7486cbb9fd81ab39d. This behaviour is really wired indeed.

I also tried for 20 times and it turned out `--standard-json` does not work on ci in the first place:
```console
$ solc@0.4.17 test /home/travis/build/watilde/solc-js
> echo {} | ./solcjs --standard-json
Empty input was read
```

Here is the actual log:
https://travis-ci.org/watilde/solc-js/jobs/281608744#L698-L702


I will have a look at `solc-bin`.

axic: Probably there is no way around changing solcjs to be asynchronous, because that hack to read stdin may or may not work. Would be nice finding a hack so it can be incrementally developed, i.e. fixing this PR and than changing solcjs entirely to be asynchronous.

tzapu: hey guys, added a fix here https://github.com/ethereum/solc-js/pull/229
seems to do what is needed
any chance you could take a look?
cheers

axic: I've pushed it here but it doesn't seem to fix it :(

tzapu: hi, thanks for trying, you most likely need to remove https://github.com/ethereum/solc-js/blob/93292b68f54ec76ae4339de6ae6c68b6cb029c22/solcjs#L58-L62 

here is my build of it `https://travis-ci.org/tzapu/solc-js/builds/394852856`, the fail was because of coveralls, as i don t have it setup/access

axic: So apparently `readFileSync(0)` works from node.js 5.x, but we still support (for some reason) 4.x

axic:
Lets merge #260 first and see if this works on all node versions.

test/cli.js
axic: Unfortunately this simple solutions fails on 4.x :(

tzapu: damn it :( i m out of ideas

axic: I'm trying to get rid of Node 4.x, unless there is a compelling reason to keep it. Please check #263

solcjs
```diff
@@ -51,15 +51,14 @@ function abort (msg) {
 }
 
 if (argv['standard-json']) {
-  var size = fs.fstatSync(process.stdin.fd).size;
-
-  if (size <= 0) {
-    abort('Empty input was read');
+  if (!solc.supportsStandard) {
```
chriseth: Do we still have that flag?

solcjs
```diff
@@ -51,15 +51,14 @@ function abort (msg) {
...
+    abort('Compiler does not support Standard JSON I/O');
   }
 
-  var input = fs.readSync(process.stdin.fd, size)[0];
+  // This accepts integer as a parameter, where 0 corresponds to stdin_fileno
+  var input = fs.readFileSync(0);
 
-  console.log(solc.compileStandardWrapper(input));
+  console.log(solc.compileStandard(input.toString('utf8')));
```
chriseth: Why not `compile`?

leonardoalt:
@chriseth updated

axic:
Didn't remove Node 4 from circleci :)

PR #128, opened 2017-07-25
Raise error message if writing a file failed

axic:
Part of #109.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -95,14 +95,27 @@ if (!output) {
 }
 
 fs.ensureDirSync (destination);
+
+function writeFile (file, content) {
+  file = path.join(destination, file);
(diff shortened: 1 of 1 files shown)


solcjs
```diff
@@ -95,14 +95,27 @@ if (!output) {
 }
 
 fs.ensureDirSync (destination);
+
+function writeFile (file, content) {
+  file = path.join(destination, file);
+  fs.writeFile(file, content, function (err) {
+    if (err) {
+      console.log('Failed to write ' + file + ': ' + err);
```
chriseth: Can we just use `console` here?

chriseth: Ah right, we  already do in other places :-)

PR #129, opened 2017-07-25
Use stderr for errors in solcjs

### LINKED ISSUE #104: Clarify error reporting in solcjs (stderr/stdout/solidity warnings vs. errors)
Right now in `solcjs` we have different outputs:
- Commandline validation errors are on stderr
- Node.js errors (e.g. file reading error) is on stderr
- Warnings & Errors from the Solidity compiler is on stdout
- In `---standard-json` mode the exit value reflects what happened
- Otherwise it is always successful

I think we should go for:
- all errors on stderr
- 'Error' from Solidity is on stderr
- 'Warning' from Solidity is on stdout
- exit(1) if stderr is not empty
- exit(0) otherwise

axic:
Fixes #104.

Credits to @chevdor for this.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -47,7 +47,7 @@ var files = argv._;
 var destination = argv['output-dir'] || '.'
 
 function abort (msg) {
-  console.log(msg || 'Error occured');
+  console.error(msg || 'Error occured');
   process.exit(1);
 }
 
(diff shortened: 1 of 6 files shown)


axic:
This is ready now and we'd need it for emscripten improvements.

chriseth:
Please rebase.

PR #141, opened 2017-09-17
Fix broken link

dongsam:
Fix changed `StandardToken.sol` link of consensys repo


axic:
@dongsam thanks! Since this code of the DAO is only included for testing and is not the original source I'd be inclined to say we shouldn't deviate too much.

Closing this now, please open a PR to the DAO to fix it.

Issue #143, opened 2017-09-22
Can't find ./bin


captDaylight:
I'm seeing this line in the README:

> Note: version strings must match the version substring of the files available in /bin/soljson-*.js

I noticed this because I'm getting an this error: 

```
./~/solc/wrapper.js
Module not found: Can't resolve './bin' in '/node_modules/solc'
```
Is there something missing from the build? I don't have a bin folder in my package.

axic:
That piece of code seems to be slightly outdated. You could use `setupMethods(<loaded file>)` instead.

captDaylight:
Thanks for the quick reply @axic. What would be the "loaded file"? I'm getting this error when I simply try to import the package, so I'm not sure how to get around it. 

Any thoughts on a fix? I could submit a PR.

axic:
I'm not sure why `useVersion` would be called, it is not used anywhere in the project. What do you mean by `import`? Node `require`? ES6 `import`?

axic:
Possibly this is the same issue as #110 ?

captDaylight:
This is happening when all I do is `require` solc into a file. The context I'm compiling with webpack and trying to get solc in the browser.

axic:
Webpack must be doing something weird. I guess it may be calling every function and instrumenting the call graph to see what is being used? Alternatively it does something with it because it has a `require` call.

gbalabasquer:
I'm having the same issue when installing the package to a react project.
`Can't resolve './bin' in '/.../node_modules/solc'`
If I comment the `useVersion` function I get now this problem:
`Module not found: Can't resolve 'module' in '/.../node_modules/require-from-string'`
In this case I comment that module and `loadRemoteVersion` function.

After commenting out those lines, it does not fail anymore at the loading part. However when using `solc.compile` I get this error:
`soljson.cwrap is not a function`. Basically because `require('./soljson.js'))` brings an empty object. I've been reading other threads about these errors but did not find a concrete solution.

ch4ot1c:
When compiling my .ts file (latest Angular, latest solc, trying to use it from within the browser via require('solc')):

ERROR in ./node_modules/require-from-string/index.js
Module not found: Error: Can't resolve 'module' in '/Users/Jon/xx/xx/artist-frontend/node_modules/require-from-string'
resolve 'module' in '/Users/Jon/xx/xx/artist-frontend/node_modules/require-from-string'
  Parsed request is a module
[...]

PR #144, opened 2017-09-22
Remove `useVersion`, no local binaries in bin 

captDaylight:
There aren't any local binaries or a bin folder for that matter so removing a method that tries to access them. I've removed `useVersion` and updated the README to reflect that.

I was getting this error when trying to simply import the solc package:

```
./~/solc/wrapper.js
Module not found: Can't resolve './bin' in '/node_modules/solc'
```

Ref issue #143

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,19 +105,7 @@ new BrowserWindow({
 
 ### Using a Legacy Version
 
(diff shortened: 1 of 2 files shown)


axic:
@chriseth I think we should merge this. `useVersion` is practically useless now since the `bin` directory is not present and due to a shortcoming in webpack this stops it from working properly.

The documentation still refers to `setupMethods` which is the viable alternative.

PR #146, opened 2017-09-25
Add stateMutability field to ABI translator

axic:
Depends on #145.


yann300:
ok but tests are failing...

axic:
It is just a lint question, will fix at rebasing after the other one is merged.

Issue #66, opened 2016-10-25
Move from semistandard to standard


axic:
This should be done after the (majority of the) current PRs are closed.

axic:
Note that `README` must be updated too, to revert changes in #390.

PR #148, opened 2017-09-30
Move from semistandard to standard

watilde:
##### Summary
+ `npm i --save-dev standard`
+ `npm rm --save-dev semistandard`
+ Use standard and fix the errors
+ Lint check at solcjs and fix the errors

Fixes https://github.com/ethereum/solc-js/issues/66.

### DIFF
--- a/abi.js
+++ b/abi.js
@@ -1,37 +1,37 @@
-var semver = require('semver');
+var semver = require('semver')
 
 function update (compilerVersion, abi) {
-  var hasConstructor = false;
-  var hasFallback = false;
+  var hasConstructor = false
+  var hasFallback = false
 
   for (var i = 0; i < abi.length; i++) {
-    var item = abi[i];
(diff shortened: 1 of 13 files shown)


watilde:
The ci error is not related to this PR.

axic:
Thanks. This is long planned, but we wanted to merge the pending pull requests first given it would be a lot of work otherwise.

watilde:
@axic That makes sense to not make the commit conflicts. Could I possibly support you to push the pending patches forward by somehow?

axic:
@watilde that would be very nice! The stdin issue is priority now, but after that #61, #63, #64 should be split into separate PRs per feature. They are useful, but hard to deal with in their current shape.

watilde:
I will close this since upstream was updated already and it's not the time to move to standard for now. Thanks!

PR #150, opened 2017-10-01
Split linker helper into its own file

axic:


linker.js
```diff
@@ -0,0 +1,26 @@
...
+
+    var hexAddress = libraries[libraryName];
+    if (hexAddress.slice(0, 2) !== '0x' || hexAddress.length > 42) {
+      throw new Error('Invalid address specified for ' + libraryName);
+    }
+    // remove 0x prefix
+    hexAddress = hexAddress.slice(2);
+    hexAddress = Array(40 - hexAddress.length + 1).join('0') + hexAddress;
+
+    while (bytecode.indexOf(libLabel) >= 0) {
```
watilde: This is including position 0, so it might be better to check with `!== -1`.

axic: 0 is a valid position in theory (it could start at the beginning), though considering EVM it must be preceded with a PUSH opcode, but I wouldn't restrict this piece of code.

Also note this code already exists in the repo, this PR only moves it.

watilde: Ahhh I see. Then that should be fine, thanks!

PR #153, opened 2017-10-13
Validate release hash on download

axic:


package.json
```diff
@@ -52,5 +53,8 @@
     "ignore": [
       "soljson.js"
     ]
+  },
+  "solc": {
+    "sha256": "a52d6dddca6c3df5a6364cc11f93a6dd1c58e38e8c740f27e39f98e43cdaafce"
```
axic: Alternatively this could come from solc-bin (there was a patch for it ages ago).

axic:
This actually triggers the reason I wanted to add this. Fails on travis:
```
Downloading correct solidity binary...
Retrieving available version list...
Downloading version soljson-v0.4.17+commit.bdeb9e52.js
Hash mismatch: 0x53d76fdde235668fe06b519bc6378aa6c8eea249fefb9bf84232801bb8c5dcdc vs 0x61664f0e2e2e1af37d369013295c25dde22416d297c7ffccba91985f3b4d033b
```

The correct hash is 0x53d.. that is what I also get locally. Sometimes (a lot of times lately) Travis kept failing on Solidity. It fails here now. Maybe some issue with new lines?

axic:
Forcing to binary encoding seems to make it work. Perhaps it will fix the majority of Travis issues we're having in Solidity.

PR #154, opened 2017-10-13
Do not treat warnings as errors in solcjs

axic:


axic:
This should fix the pre-release issue and Solidity tests.

solcjs
```diff
@@ -89,8 +89,17 @@ var output = solc.compile({ sources: sources }, argv.optimize ? 1 : 0);
 if (!output) {
   abort('No output from compiler');
 } else if (output['errors']) {
+  function isWarning (message) {
+    return message.match(/^(.*:[0-9]*:[0-9]* )?Warning: /)
```
axic: This is the regexp from browser-solidity so it is "battle tested". Eventually this entire code should move onto jsonio alleviating this problem.

solcjs
```diff
@@ -89,8 +89,17 @@ var output = solc.compile({ sources: sources }, argv.optimize ? 1 : 0);
...
+  }
+
   for (var error in output['errors']) {
-    console.error(output['errors'][error]);
+    var message = output['errors'][error]
+    if (isWarning(message)) {
+      console.log(message)
```
chriseth: Does solcjs produce any other output on stdout? My feeling is that the data should go to stdout and anything user-facing through the "side-channel" stderr. But if we do not have output on stdout anyway (I didn't see anything), then this is fine.

axic: Data goes into files.

axic:
I'm merging this for the moment, but we should probably improve this repo a lot.

PR #157, opened 2017-10-17
WIP: trying to fix download

axic:


axic:
It seems to be working now. Probably it was a travis hiccup.

PR #159, opened 2017-10-19
Use outputSelection in jsonio tests

axic:


axic:
@chriseth can this be merged?

chriseth:
Wouldn't it be better to request everything in most tests and only have some test that only request the bytecode?

axic:
I had `'*'` before but that is not included in the docs (yet) and all the tests only care about `evm.bytecode` so I've figured the tests will be faster this way (once implemented properly).

chriseth:
ok sure, but do we at least have tests that request everything?

axic:
Not here, we have them in Solidity.

Issue #135, opened 2017-08-19
Solc module not found errors after npm installing solc 


sivachaitanya:
I have installed solc using "npm install solc" in my project directory, however after refreshing my webpage I see a bunch of Module not found related errors,following is my web3js code 
       

>  ecrecTestSource = "big string of my contract";
>          ecrecTestCompiled  = solc.compile(ecrecTestSource, 1);
>       
>         ecrecTestCode = ecrecTestCompiled.contracts['EcrecTest'].bytecode;
>         ecrecTestABI =  JSON.parse(ecrecTestCompiled.contracts['EcrecTest'].interface);
>         ecrecTestContract = web3.eth.contract(ecrecTestABI);
>         deployedecrecTest = ecrecTestContract.at(contractAddress);

The following errors are seen on my webpage, can somebody please help me resolve this issue

![chrome_2017-08-19_21-34-58](https://user-images.githubusercontent.com/6172334/29488442-e11fd240-8528-11e7-9cab-588bcfeb8190.png)

axic:
Would you mind sharing your code? It seems you are not loading the code correctly.

sivachaitanya:
```
import "../stylesheets/app.css";
import { default as Web3} from 'web3';
import { default as contract } from 'truffle-contract';
 const util = require('ethereumjs-util');
 var HookedWeb3Provider = require("hooked-web3-provider");
 var lightwallet = require('eth-lightwallet');
 var signing = require('eth-lightwallet/lib/signing');
 var solc = require('solc');
 const fs = require('fs');
 var keystore= lightwallet.keystore;
 var ks_temp;
 var seed;
 var pvtkey;
 var pwDerivedKey_store;
var accounts_node;
var account;
var address = "0x692a70d2e424a56d2c6c27aa97d1a86395877b3a";
var web3;
var ecrecTestABI;
var ecrecTestContract;
var ecrecTestCode;
var deployedecrecTest;
var org_index;
window.App = {
  start: function() {
    var self = this;
    web3.eth.getAccounts(function(err, accs) {
      if (err != null) {
        alert("There was an error fetching your accounts.");
        return;
      }
      if (accs.length == 0) {
        alert("Couldn't get any accounts! Make sure your Ethereum client is configured correctly.");
        return;
      }
      console.log(accs);
      accounts_node = accs;
      account = accounts_node[0];
      web3.eth.defaultAccount= account;
[...]

Issue #110, opened 2017-05-26
Please provide commonJS entry point for solc-js


Tectract:
Can't for the life of me get solc working in a react app:

I need to be able to do this:

import solc from 'solc';

but It's failing right now. Can't get BrowserSolc to load under react either. And Geth has removed eth_compileSolidity...

So, it would be wonderful if solc had a commonJS entry point. I'm not sure how to add one though.

axic:
Not entirely sure what this issue means, since as far as I know CommonJS defines that `module.exports` refers to what a module exports and that is being done here.

Tectract:
I ended up using browser-solidity, a minimized tool which autoloads the latest sold compiler into the browser, and doing compilation with that. Being able to just do:

`import solc from solc`

would have been really nice. Moot point now, see how I'm doing browser-side compilation here:

https://github.com/Tectract/EthDeployer/blob/master/src/Deploy.js

lukehedger:
@Tectract The CommonJS module is exposed by default and can be used in Node.js or module bundlers (like Webpack) with `const solc = require('solc')`. To be honest though, Webpack should also be able to resolve the dependency with `import` syntax.

Otherwise, if you want to use the ES6 Module syntax `import solc from 'solc'` you'll need `solc-js` in ES6 Module format, rather than CommonJS.

@axic The simplest way to compile a JS library to both CommonJS and ES6 modules is to use [Rollup](https://github.com/rollup/rollup).

spirosbax:
The easiest way to use solc-js with webpack on the browser right now is:
`const solcWrapper = require('solc/wrapper')`  
`const solc = solcWrapper(window.module)`

Tectract:
>  if you want to use the ES6 Module syntax import solc from 'solc' you'll need solc-js in ES6 Module format, rather than CommonJS

Ok, interesting. I'm still learning about how these package systems work I guess. Is there an ES6 module version of solc available?

spirosbax:
> > if you want to use the ES6 Module syntax import solc from 'solc' you'll need solc-js in ES6 Module format, rather than CommonJS
> 
> Ok, interesting. I'm still learning about how these package systems work I guess. Is there an ES6 module version of solc available?

as far as i know, no.
In my case using that wrapper would sometimes cause errors, so I stopped using solc.

matheusaaguiar:
Closing for inactivity.

PR #162, opened 2017-11-17
require-from-string from ^1.1.0 to 1.2.0

ch4ot1c:
#110 #135 #143 

Tried to use `solc` in a browser with Angular (latest), but was getting an error at buildtime. This solves the problem for me.

Here is the commit in question (from their repo): https://github.com/floatdrop/require-from-string/commit/80783b2ca46daa4157241503e434f10a24bc4c67

Ran the `solc` tests. All the initial ones passed; as it got to the end there were several ENOENTS with DAO files. Perhaps I didn't fetch everything... but in general it passes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "dependencies": {
     "fs-extra": "^0.30.0",
     "memorystream": "^0.3.1",
-    "require-from-string": "^1.1.0",
+    "require-from-string": "1.2.0",
     "semver": "^5.3.0",
     "yargs": "^4.7.1"
   },


ch4ot1c:
Bah - I jumped the gun. New error in Chrome console:
```
Invalid asm.js: Invalid member of stdlib
vendor.bundle.js:32806 Uncaught TypeError: soljson.cwrap is not a function
    at setupMethods (vendor.bundle.js:32806)
    at Object.../../../../solc/index.js (vendor.bundle.js:32606)
    at __webpack_require__ (inline.bundle.js:55)
    at Object.../../../../../src/app/app.component.ts (main.bundle.js:46)
    at __webpack_require__ (inline.bundle.js:55)
    at Object.../../../../../src/app/app.module.ts (main.bundle.js:495)
    at __webpack_require__ (inline.bundle.js:55)
    at Object.../../../../../src/main.ts (main.bundle.js:559)
    at __webpack_require__ (inline.bundle.js:55)
    at Object.0 (main.bundle.js:577)
```
@chriseth I looked at `browser-solidity` as an option, but that looks like the Web IDE. Trying to do simple JIT compilation & deployment of a .sol file, from a dApp (from the client).

My familiarity with webpack is average, so I'm not sure how much further I can go down this rabbit hole.

package.json
```diff
@@ -39,7 +39,7 @@
   "dependencies": {
     "fs-extra": "^0.30.0",
     "memorystream": "^0.3.1",
-    "require-from-string": "^1.1.0",
+    "require-from-string": "1.2.0",
```
axic: Is there a good reason for fixing it to `1.2.0`? Can we just go with `^1.2.0`?

ch4ot1c: Unsure - ended up using a constructor, so didn't need browser solc and the change in this PR. False alarm.

I think ^1.2.0 is ok, but I haven't tested it.

StefH:
Any update on this one?

axic:
Relevant issue is here: https://github.com/floatdrop/require-from-string/issues/18

chriseth:
@axic now they do: https://github.com/ethereum/solc-js/issues/576

PR #167, opened 2017-12-05
Use stringToUTF8 instead of deprecated writeStringToMemory

axic:


axic:
This is needed to avoid warnings on soljson builds on newer emscripten versions. The circleci target of Solidity uses a newer emscripten which triggers this problem.

Apparently this method is available since a few versions earlier than 1.35 the regular Solidity build system uses.

According to https://groups.google.com/forum/#!topic/emscripten-discuss/BoFjUXAE0D4:
- the old one was deprecated 1.36.8
- the new one existed since 1.29.4

Need to be tested though with both builds.

wrapper.js
```diff
@@ -15,7 +15,7 @@ function setupMethods (soljson) {
   if (('_compileJSONCallback' in soljson) || ('_compileStandard' in soljson)) {
     var copyString = function (str, ptr) {
       var buffer = soljson._malloc(str.length + 1);
-      soljson.writeStringToMemory(str, buffer);
+      soljson.stringToUTF8(str, buffer, str.length + 1);
```
chriseth: We have to check whether we allocate enough space for the UTF8 encoding.

axic:
Fixed the style failures.

@chriseth according to coverage the those lines are touched in 2 tests so I guess it should be safe to merge :)

wrapper.js
```diff
@@ -14,8 +14,9 @@ function setupMethods (soljson) {
   var compileStandard = null;
   if (('_compileJSONCallback' in soljson) || ('_compileStandard' in soljson)) {
     var copyString = function (str, ptr) {
-      var buffer = soljson._malloc(str.length + 1);
-      soljson.writeStringToMemory(str, buffer);
+      var length = soljson.lengthBytesUTF8(str);
+      var buffer = soljson._malloc(length + 1);
+      soljson.stringToUTF8(str, buffer, length + 1);
```
axic: Code here: https://github.com/kripken/emscripten/blob/78b44ed55cc4d0b4d79f62df9e80ae6f29a5345b/src/preamble.js#L1612-L1625

> // stringToUTF8Array always appends null.

> // Copies the given Javascript String object 'str' to the emscripten HEAP at address 'outPtr',
> // null-terminated and encoded in UTF8 form. The copy will require at most str.length*4+1 bytes of space in the HEAP.
> // Use the function lengthBytesUTF8 to compute the exact number of bytes (excluding null terminator) that this function will write.
> // Returns the number of bytes written, EXCLUDING the null terminator.

So I think the above code now is correct.

PR #168, opened 2017-12-19
Fix gas estimation translation

yann300:


translate.js
```diff
@@ -60,7 +60,15 @@ function translateJsonCompilerOutput (output) {
...
     var gasEstimates = contractInput['gasEstimates'];
-
+    if (!gasEstimates['creation']) {
+      gasEstimates['creation'] = [null, null];
+    }
+    if (!gasEstimates['internal']) {
+      gasEstimates['internal'] = null;
+    }
+    if (!gasEstimates['external']) {
+      gasEstimates['external'] = null;
```
axic: Why does it need to be set to `null`?

yann300: cause this is how `translateGasEstimates` manage it

axic: Oh, probably that shouldn't check for `null` but for `undefined`. Will need to double check.

yann300:
@chriseth @axic can we merge this?

translate.js
yann300: `translateGasEstimates` check for null

axic:
@yann300 can you check this now? The previous version was creating invalid output, e.g. `{ "internal": "infinite" }` whereas `internal` is expected to be a hashmap.

yann300:
looks good too

PR #176, opened 2018-01-11
Update translate.js

yann300:
@axic @chriseth


axic:
We should really have tests for these.

axic:
@yann300 sorry for not seeing this earlier

PR #177, opened 2018-02-04
Run code coverage report on coveralls

axic:

PR #178, opened 2018-02-04
Properly test fallback functions in abi translator

axic:
Depends on #177.


axic:
Depends on #179 as node 4 build fails without that.

PR #179, opened 2018-02-04
Only build on stable node version (4, 6, 8) on travis

axic:


package.json
```diff
@@ -44,7 +44,7 @@
     "yargs": "^4.7.1"
   },
   "devDependencies": {
-    "ethereumjs-util": "^5.1.2",
+    "ethereumjs-util": "5.1.2",
```
axic: For reason see https://github.com/ethereumjs/ethereumjs-util/issues/117

axic: Apparently with the next release (hopefully today) we can remove the pinning.

axic: 5.1.4 should fix node 4 support - will require that if travis succeeds.

axic: 5.1.4 should have fixed it - running travis again with that.

axic:
Works, squashing commits.

PR #190, opened 2018-02-13
Support running compiler unit tests on older (read: <0.4.9) compilers

axic:


axic:
Tested with 0.2.0, 0.3.5-nightly.2016.6.14, 0.3.6, 0.4.5 and 0.4.20-nightly.2018.2.13.

Fails with 0.4.0 which seems to be quite a broken release (internal compiler errors for a lot of things).

Still fails with the matuselah 0.1.1 :)

axic:
Works now with 0.1.1, 0.1.2, 0.1.3 and 0.1.4 too :)

axic:
Most of the tests are disabled for <0.1.6 as they require multiple file support. Solution will be to write them as single source tests in the future.

translate.js
```diff
@@ -8,7 +8,7 @@
 /// New style: 0.4.5+commit.b318366e.Emscripten.clang
 function versionToSemver (version) {
   // FIXME: parse more detail, but this is a good start
-  var parsed = version.match(/^([0-9]+\.[0-9]+\.[0-9]+)-([0-9a-f]{8})[\/\*].*$/);
```
chriseth: I don't understand the description of the commit. This is just about not escaping the special symbols, which is not required here anyway.

axic: That's it.

PR #191, opened 2018-02-13
Legacy assembly formatter

axic:


axic:
@chriseth please review

translate.js
```diff
@@ -148,7 +148,41 @@ function translateJsonCompilerOutput (output) {
   return ret;
 }
 
+function formatAssemblyText (asm, prefix, source) {
+  if (typeof asm === typeof '' || asm === null || asm === undefined) {
+    return prefix + asm + '\n';
```
chriseth: This will print `undefined` - do we want that?

translate.js
```diff
@@ -148,7 +148,41 @@ function translateJsonCompilerOutput (output) {
...
+  }
+  var text = prefix + '.code\n';
+  asm['.code'].forEach(function (item, i) {
+    var v = item.value === undefined ? '' : item.value;
+    var src = '';
+    if (source !== undefined && item.begin !== undefined && item.end !== undefined) {
+      src = source.slice(item.begin, item.end).replace('\n', '\\n', 'g');
```
chriseth: What about `\t`? Perhaps rather use some escaping function?

translate.js
axic: No. This was a verbatim copy from browser-solidity. I guess it should be `prefix + (asm || '') + '\n'`

translate.js
axic: True, fixed. Haven't found any built in escape function for those, do you know any?

axic:
@chriseth ready for a second review

axic:
@yann300 @iurimatias this can now be used in remix :)

PR #192, opened 2018-02-14
Support finding link references

### LINKED ISSUE #172: Support linkReferences in translate
See https://github.com/ethereum/browser-solidity/pull/968

axic:
Fixes #172.


README.md
```diff
@@ -135,13 +135,27 @@ solc.loadRemoteVersion('latest', function (err, solcSnapshot) {
...
-bytecode = solc.linkBytecode(bytecode, { 'MyLibrary': '0x123456...' })
+var linker = require('solc/linker')
+
+bytecode = linker.linkBytecode(bytecode, { 'MyLibrary': '0x123456...' })
 ```
 
-Note: in future versions of Solidity a more sophisticated linker architecture will be introduced.  Once that changes, this method will still be usable for output created by old versions of Solidity.
+(Note: `linkBytecode` is also exposed via `solc` as `solc.linkBytecode`, but this usage is deprecated.)
+
+As of Solidity 0.4.11 the compiler supports [standard JSON input and output](https://solidity.readthedocs.io/en/develop/using-the-compiler.html#compiler-input-and-output-json-description) which outputs a *link references* map. This gives a map of library names to offsets in the bytecode to replace the addresses at.
```
chriseth: Perhaps mention that this one works with longer paths and library names while the other one doesn't.

PR #199, opened 2018-02-25
Link bytecode if libraries were supplied in compileStandardWrapper

axic:


axic:
@chriseth please review again, this is now compatible with json io

chriseth:
Needs rebase.

PR #201, opened 2018-02-28
Ensure that library address has hex prefix in JSONIO

axic:


test/package.js
```diff
@@ -389,7 +389,7 @@ tape('Compilation', function (t) {
       'settings': {
         'libraries': {
           'lib.sol': {
-            'L': '4200000000000000000000000000000000000001'
+            'L': '0x4200000000000000000000000000000000000001'
```
axic: Finally this is following the documentation 😉

PR #205, opened 2018-03-05
Adds Typescript definitions.

MicahZoltu:
This isn't complete, but it is _way_ better than nothing and gives a place for people to add additional typescript definitions in the future.  Recommend switching over to TypeScript for this library so this file is generated automatically.  :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,156 @@
+declare module 'solc' {
+  export type Primitive =
+    'bool' |
+    'string' |
+    'address' |
+    'uint8' |
+    'uint16' |
+    'uint32' |
+    'uint64' |
+    'uint128' |
+    'uint256' |
+    'int8' |
+    'int16' |
+    'int32' |
+    'int64' |
+    'int128' |
+    'int256' |
+    'bytes' |
+    'bytes20' |
+    'bytes32' |
+    'bool[]' |
+    'string[]' |
+    'address[]' |
+    'uint8[]' |
+    'uint16[]' |
+    'uint32[]' |
+    'uint64[]' |
+    'uint128[]' |
+    'uint256[]' |
+    'int8[]' |
+    'int16[]' |
+    'int32[]' |
+    'int64[]' |
+    'int128[]' |
+    'int256[]' |
+    'bytes[]' |
+    'bytes20[]' |
+    'bytes32[]';
+
+  export interface AbiParameter {
+    name: string,
+    type: Primitive,
+  }
+  
+  export interface AbiEventParameter extends AbiParameter {
+    indexed: boolean,
+  }
+  
+  export interface AbiFunction {
+    name: string,
+    type: 'function' | 'constructor' | 'fallback',
+    stateMutability: 'pure' | 'view' | 'payable' | 'nonpayable',
+    constant: boolean,
+    payable: boolean,
+    inputs: Array<AbiParameter>,
+    outputs: Array<AbiParameter>,
+  }
+  
+  export interface AbiEvent {
+    name: string,
+    type: 'event',
+    inputs: Array<AbiEventParameter>,
+    anonymous: boolean,
+  }
+  
+  export type Abi = Array<AbiFunction | AbiEvent>;
+  
+  interface CompilerInputSourceFile {
+    keccak256?: string;
+    urls: string[];
+  }
+  interface CompilerInputSourceCode {
+    keccak256?: string;
+    content: string;
+  }
+  interface CompilerInput {
+    language: "Solidity" | "serpent" | "lll" | "assembly";
+    settings?: any,
+    sources: {
+      [globalName: string]: CompilerInputSourceFile|CompilerInputSourceCode,
+    };
+  }
+  interface CompilerOutputError {
+    sourceLocation?: {
+      file: string;
+      start: number;
+      end: number;
+    };
+    type: "TypeError" | "InternalCompilerError" | "Exception";
+    component: "general" | "ewasm";
+    severity: "error" | "warning";
+    message: string;
+    formattedMessage?: string;
+  }
+  interface CompilerOutputEvmBytecode {
+    object?: string;
+    opcodes?: string;
+    sourceMap?: string;
+    linkReferences?: {} | {
+      [globalName: string]: {
+        [name: string]: {start: number, length: number}[];
+      };
+    };
+  }
+  interface CompilerOutputSources {
+    [globalName: string]: {
+      id: number;
+      ast?: any;
+      legacyAST?: any;
+    },
+  }
+  interface CompilerOutputContracts {
+    [globalName: string]: {
+      [contractName: string]: {
+        abi?: Abi;
+        metadata?: string;
+        userdoc?: any;
+        devdoc?: any;
+        ir?: string;
+        evm?: {
+          assembly?: string;
+          legacyAssembly?: any;
+          bytecode: CompilerOutputEvmBytecode;
+          deployedBytecode?: CompilerOutputEvmBytecode;
+          methodIdentifiers?: {
+            [methodName: string]: string;
+          };
+          gasEstimates?: {
+            creation: {
+              codeDepositCost: string;
+              executionCost: string;
+              totalCost: string;
+            };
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,156 @@
...
+    [globalName: string]: {
+      id: number;
+      ast: any;
+      legacyAST: any;
+    },
+  }
+  interface CompilerOutputContracts {
+    [globalName: string]: {
+      [contractName: string]: {
+        abi: Abi;
```
axic: All of these are optional. I'm not sure it is a good idea to have forced description the JSON, it is not actually part of solc-js, but part of Solidity.

MicahZoltu: You mean `abi` and `evm`?  The rest are already marked as optional.

MicahZoltu:
> I'm not sure it is a good idea to have forced description the JSON, it is not actually part of solc-js, but part of Solidity.

As a user, I depend on solcjs, not Solidity (I get that transitively). Therefore, I should be able to get the typescript definitions through solcjs.  Specifically, I call functions in the `solcjs` library and it takes parameters/returns values that have a type and if I am using TypeScript those types should be exposed to me without needing to include some additional library.

I don't believe there is any easy way to bundle TypeScript definitions with Solidity such that I can get them transitively via solcjs, which is why I have added them here.  I'm certainly open to other mechanisms for making the definition files available, but this seemed like the best option from a usability standpoint.

index.d.ts
axic: Everything in the JSON output is optional as it depends on the JSON input.

MicahZoltu: Is there anywhere that this is documented?  When you say "everything is optional" I'm pretty sure you don't actually mean everything as (for example) I don't believe that you can configure the output to omit `type` in the `AbiFunction` structure.  Perhaps you just mean "each of the top level keys under `[contractName: string]` are optional"?  Are all of them actually optional or are some of them optional in the sense that they may be empty objects or empty arrays?  Looking at the docs, the example shows several as having empty arrays or empty objects and I _vaguely_ remember when I was working with these a while back encountering some situations where rather than being omitted an empty object or array was included (but I could be misremembering).

My goal here is to accurately represent the structure of the object returned by the Solidity compiler.  My inquiries are because I really have no idea exactly what it _can_ spit out, so I need some help specifying it.

If you really want to go above and beyond to help out, a comment on each non-optional thing that can be optional would be ideal.  😄

MicahZoltu: Looking over the docs again, I have added a few more optional indicators (locally).  However, it is unclear from the docs whether `evm.bytecode` will be included if all of the members of `evm.bytecode` are excluded?  Or perhaps it will just be an empty object in this case?  Similar question for `ewasm` section, is it an empty object if `ewasm.wast` and `ewasm.wasm` are filtered out or is it omitted entirely if they are filtered out?

MicahZoltu: Same question for the top level things, will the `contracts` key be omitted entirely if everything is filtered out from it, or will it always be present with an empty object?

axic: All of it is documented here: http://solidity.readthedocs.io/en/develop/using-the-compiler.html#compiler-input-and-output-json-description

The ABI is documented here (but that is again outside the scope of the JSON compiler interface, it is a specification shared between languages and clients): http://solidity.readthedocs.io/en/develop/abi-spec.html#json

MicahZoltu:
The two links you provided (@axic) in your comment how I created this file in the first place.  If you believe I have mis-interpreted the documentation or there is otherwise an error in this definition I encourage you to point it out (please be specific).  I have updated the PR with a couple changes for where I believe I had a mistake previously (non-optional fields), but the documentation is not particularly clear on what happens when a node in the object graph would be an empty array or empty object, will the node be omitted or will it be empty?

MicahZoltu:
Examples of what is unclear are the top three keys (errors, sources, contracts) and `contracts.[globalName].[contractName].evm`, `.ewasm`, and `.evm.bytecode`.

danqing:
Any progress on this? Would love to have TypeScript defs.

cruhl:
It seems like this is ready to go... any updates? Thanks in advance!

axic:
@MicahZoltu @danqing @cruhl sorry for the delays, we just need to understand Typescript better since otherwise we cannot maintain it and merging means we have to maintain it.

Would it be possible to add tests which use the typescript definition?

index.d.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+  interface CompilerOutputContracts {
+    [globalName: string]: {
+      [contractName: string]: {
+        abi?: Abi;
+        metadata?: string;
+        userdoc?: any;
+        devdoc?: any;
+        ir?: string;
+        evm: {
```
axic: Even `evm` is optional, not only the fields below. Practically everything is optional in the response JSON.

MicahZoltu: Fixed.

MicahZoltu: Side note: I'm not a fan of responses where everything is optional, it makes the receiver have to litter their code with null coalescing which is painful in JavaScript.  However, that is out of scope of this PR which is just trying to match reality.

andys8:
Is this definition also relevant for the user with `solc/wrapper`?

I'm trying to match the definitions with the actual use or the code in the `README.md`. E.g. this line:

```js
console.log(contractName + ': ' + output.contracts[contractName].bytecode)
```

But `CompilerContracts` doesn't have a field `bytecode` on top level.
https://github.com/ethereum/solc-js/blob/bab0fef1eae729610421344378bfcb0f65e152f3/index.d.ts#L112-L148

MicahZoltu:
Last I checked `contracts[contractName]` doesn't have a `bytecode` field on it directly.  Perhaps this has changed?

andys8:
Looking at the Readme of the repository and the definitions I would assume it will not work. Also it did not work for me in a local project with this specific example.

As far as I know, the Typescript people write tests which work with their definitions and run the Typescript compiler on them. Could be an option.

MicahZoltu:
At the point in time when I submitted this PR it worked.  `npm install solc` and dump this in your project's definitions folder.  This project compiles and runs: https://github.com/AugurProject/augur-core/blob/2d9cbac69e33be2f07a02ad91fe51574994be066/source/libraries/ContractCompiler.ts#L154

sci4me:
Is this still being worked on? Would love to have TypeScript definitions for solc-js!

MicahZoltu:
I'm not actively working on it, but would still like to get it merged.  I'm unclear if there is anything realistic that we could do to get this merged.  To maintainers: Can you decide yay or nay on this rather than limbo?  If you decide nay it can be added to DefinitelyTyped so people can still benefit (though not as cleanly).

ToJen:
Any update on this PR?

axic:
The API with 0.5.0 changed significantly. It only supports strings externally and as a result there's no need to define the JSON (including ABI JSON).

@MicahZoltu can you update the PR? I'd like to merge it after.

axic:
To clarify: if you could create a PR which only adds types to the public methods on wrapper (e.g. without the standard JSON/ABI) that could be merged quickly. The rest of this PR is more complicated.

MicahZoltu:
Hmm, I haven't followed Solidity 5.0, and I don't understand what you mean by "it only supports strings externally".

MicahZoltu:
I have created ea new PR that only contains the `compile` callback definition: https://github.com/ethereum/solc-js/pull/298

While I'm happy to see that PR get merged as it is a good starting point for getting TypeScript definitions included, I don't feel that it is _nearly_ as useful as getting the JSON input/output well defined.

I'm also curious why the `compile` function takes a string and returns a string when what it really wants/provides is a very specifically defined JSON object.  In normal JS the difference is minimal, but in TypeScript the difference is pretty huge as you can get strong type checking and hinting if it takes a well defined object while "string" results in basically no type checking.

r0qs:
Hi, thank you very much for this PR. However, part of what was discussed here was already implemented or is under development in other PRs (e.g. https://github.com/ethereum/solc-js/pull/630), so I will be closing this for now.

Please, feel free to reopen it if necessary.

PR #208, opened 2018-03-28
Fix linkBytecode to work properly with two-level lists

theoport:
When using standard JSON i/o, input parameter `libraries` to function `linkBytecode` is of form
```
{
"myFile.sol": {
        "MyLib": "0x123123..."
      }
}
```
I believe the typeof check is meant to be applied to the value not the the key of `libraryName`


axic:
Thanks! I wonder why the tests were passing. Can you try extending the test suite to catch this problem?

theoport:
Sure, will see if I have time on the weekend

theoport:
@axic

axic:
Thanks! I've realised there's much else to fix, but will do that separately.

PR #210, opened 2018-04-15
Implements solc-latest feature for nightly builds

ProphetDaniel:
This implements latest solc feature and can be used to test nightly builds.


ProphetDaniel:
Actually this shouldn't be merged to master branch but a new solc-latest branch should actually be created to handle this functionality.

axic:
Can you please explain what is the goal of this PR?

ProphetDaniel:
@axic, the goal is to have a packaged solc node module updated once every nightly build, which would be automatically triggered by nightly builds. So this one downloads the latest nightly version instead of package.json indicated release version.

ProphetDaniel:
Well thought!

r0qs:
Hi, thank you very much for the PR, and sorry for the late reply.
But I will be closing this PR now since this feature is already supported in the current version.

You can use it like so:

```javascript
solc.loadRemoteVersion("v0.8.15-nightly.2022.5.27+commit.095cc647", (err, compiler) => {
    console.log(compiler.version())
})
```

Issue #219, opened 2018-06-10
CLI generates weird names for smart constracts files


krzkaczor:
Command: `solcjs --overwrite --abi --bin -o ./test/contract-artifacts ./test/contracts/*`

When using native `solc`, generated files have reasonable names like `DumbContract.abi`, `DumbContract.bin` (same as original). With solcjs filenames contain full path like: `__test_contracts_DumbContract_sol_DumbContract.abi`

Is this behaviour intended? It makes it painful to work with generated files. We do it in integration tests for one of my projects and its super weird to work with these files later: https://github.com/Neufund/TypeChain/blob/master/test/integration/DumbContract.spec.ts#L5

PR #220, opened 2018-06-10
Change rules for selecting filename for generated files

krzkaczor:
Closes https://github.com/ethereum/solc-js/issues/219.
Behaviour is now the same as for native `solc`.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -115,7 +115,7 @@ function writeFile (file, content) {
 }
 
 for (var contractName in output.contracts) {
-  var contractFileName = contractName.replace(/[:./]/g, '_');
+  var contractFileName = contractName.replace(/(.*):/, '');
 
   if (argv.bin) {
     writeFile(contractFileName + '.bin', output.contracts[contractName].bytecode);


solcjs
```diff
@@ -115,7 +115,7 @@ function writeFile (file, content) {
 }
 
 for (var contractName in output.contracts) {
-  var contractFileName = contractName.replace(/[:./]/g, '_');
+  var contractFileName = contractName.replace(/(.*):/, '');
```
chriseth: What happens in case of collisions?

chriseth:
solc is looking for collisions. If there are none, it uses the contract name, otherwise the fully qualified name with `:` and others replaced by `_`

krzkaczor:
@chriseth okay i can recrete this behaviour in solcjs. I just think that current state of things is wrong.

r0qs:
Hi, I believe we can have this. And maybe add a `--overwrite` flag like the one in `solc` to prevent the name collisions? So the behavior of solc-js and solc would be compatible.
What do you think @chriseth?

@krzkaczor would you like to finish this or should we take it over?

krzkaczor:
@r0qs please, take over :)

PR #221, opened 2018-06-12
Update DAO contracts.

ekpyron:
I created a fork, since I don't have write permissions for the repository.

So far this is entirely untested.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -1,6 +1,14 @@
 const tape = require('tape');
 const spawn = require('tape-spawn');
 const pkg = require('../package.json');
+const semver = require('semver');
+
+var daodir;
+if (semver.lt(pkg.version, '0.5.0')) {
+  daodir = 'DAO040';
+} else {
+  daodir = 'DAO';
+}
 
 tape('CLI', function (t) {
   t.test('--version', function (st) {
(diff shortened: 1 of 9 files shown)


test/cli.js
```diff
@@ -1,6 +1,15 @@
 const tape = require('tape');
 const spawn = require('tape-spawn');
 const pkg = require('../package.json');
+const solc = require('../index.json');
```
axic: This file is using the `CLI`, it shouldn't acquire the version through this I think. It should check `./solcjs --version`.

axic:
The rest looks good.

test/determinism.js
```diff
@@ -1,16 +1,23 @@
...
 
 tape('Deterministic Compilation', function (t) {
   t.test('DAO', function (st) {
     var input = {};
     var prevBytecode = null;
+    var testdir;
+    if (semver.lt(solc.semver(), '0.5.0')) {
+      testdir = 'test/DAO040/';
+    } else {
+      testdir = 'test/DAO/';
```
axic: Worth mentioning that due to the current `develop` branch of solidity not using `0.5.0`, this change on its own will not be enough to make it pass. However we can just delete the `DAO040` directory an symlink `DAO` to it in our test suite.

test/cli.js
ekpyron: Hm - if I understand this file correctly, the first test checks whether ``./solcjs --version`` matches ``pkg.version`` anyways, so it may make sense to use ``pkg.version`` to choose the DAO directory? However ``pkg.version``, resp. ``version`` in ``../package.json`` is fixed to ``0.4.24``... I'm actually not sure how this will be run for different versions in the end.

test/determinism.js
ekpyron: Yes, I was considering using 0.4.25 instead, but using 0.5.0 probably makes more sense, but we have to deal with this somehow.

PR #223, opened 2018-06-12
Update DAO contracts (use strict address literals)

axic:
Some of these were missed in 40151cc. Fixes #222.


axic:
Pulling the plug here because this isn't a substantial change to the library, only the DAO test case.

PR #224, opened 2018-06-14
Update test code to support new way of using keccak256 and call.

chriseth:


chriseth:
Required for https://github.com/ethereum/solidity/pull/4097

PR #227, opened 2018-06-19
Add solcjs-get

chevdor:
```
$ ./solcjs-get --help
Usage: [options| version]

Options:
  --version   Show version number                                      [boolean]
  --list      List all the versions
  --all       Download all the versions. Additionnaly, use --nightly to get the
              nightly builds.
  --nightly   Combined with --all, it adds the nightly to the list
  --releases  Get all the release builds
  --clean     Delete all the compiler in ./bin
  --help      Show help                                                [boolean]
```


chevdor:
It replaces `downloadCurrentVersion.js`

axic:
> It replaces downloadCurrentVersion.js

In the PR it doesn't do that yet. I can look into fixing it if you don't have time (which I understand, given how long it took to get into reviewing the PR).

chevdor:
I admit I lost the history of this one...
I initially made a big PR (very likely *too* big). Upon your request I then shrunk it a bit so parts may have been left out intentionally. Feel free to fix it as you wish but keep an eye on the other PRs as your fixes may be already done elsewhere.

chevdor:
Hello, coud this PR be reviewed and either merged or closed ?

chriseth:
Can you explain the benefit over `downloadCurrentVersion.js`?

chevdor:
> Can you explain the benefit over downloadCurrentVersion.js?

Back in 2018 I surely could have 😄 
From what I recall, instead of downloading ONLY the current version, it allowed the user downloading the version of their choice.

cameel:
I think this looks useful. It's at least an improvement over the original script.

I rebased it on the latest `master` and added some small tweaks (typos, parameter descriptions, style). I also modified so that the new script actually replaces `downloadCurrentVersion.js`.

There are some things that I think get unnecessarily removed. I'll point them out in review.

solcjs-get
```diff
@@ -4,10 +4,39 @@
...
+    type: 'bool'
+  })
+  .option('releases', {
+    describe: 'Get all the release builds',
+    type: 'bool'
+  })
+  .option('clean', {
+    describe: 'Delete all the compiler in ./bin',
+    type: 'bool'
+  })
```
cameel: These options all seem to be mutually exclusive but the script does not print any error if you use more than one. This was pretty confusing when I tried running it for the first time. Using more than one should not be allowed unless it actually does something.

BTW, I think that combining `--list` with `--all`/`--nightly`/`--releases` might actually be useful. I'd even make `--list` print only releases by default because the whole list is pretty long.

solcjs-get
```diff
@@ -25,51 +54,83 @@ function getVersionList (cb) {
...
+      wanted = requestedVersion;
+    } else {
+      console.log('Requested version: latest release');
+      wanted = list.releases[pkg.version.match(/^(\d+\.\d+\.\d+)$/)[1]];
+    }
+
+    downloadBinary(wanted, function (file) {
+      if (!requestedVersion){
+        fs.copy(file, 'soljson.js'); // for backward compatibility
+      }
```
cameel: This should really be a symlink (at least if the FS supports it).

Actually, this does not seem like something just for backwards compatibility. Can solc-js even use the compiler when it's inside `bin/`? I think that an option to relink it to another version might be useful because that's basically how the select the compiler version now.

solcjs-get
```diff
@@ -25,51 +54,83 @@ function getVersionList (cb) {
...
+    process.exit(0);
+  });
+} else if (argv.clean) {
+  console.log('Removing all local compilers in ' + compilerDir + ' ...');
+  fs.remove(compilerDir, function (err) {
+    if (err) return console.error(err);
+
+    console.log('Success! Cleaned ' + compilerDir);
+    process.exit(0);
+  });
```
cameel: Not sure if it's a good idea to remove the whole dir. As a user I think I'd expect this to be more intelligent and only remove the binaries (otherwise why even provide an option for something that can be easily done with `rm`?). I think it would be safer to only remove files matching the valid name regex.

solcjs-get
```diff
@@ -25,51 +54,83 @@ function getVersionList (cb) {
...
+} else if (argv.all) {
+  console.log('Getting all the versions ...');
+  getVersionList(function (list) {
+    list = JSON.parse(list).builds;
+    for (var i = list.length - 1; i >= 0; i--) {
+      var target = list[i].path;
+      if (target.indexOf('nightly') > 0 && !argv.nightly) continue;
```
cameel: The list contains a `releases` array. I think that checking if the file name is not in it would be a more future-proof check for a nightly.

solcjs-get
```diff
@@ -4,10 +4,39 @@
...
-var fs = require('fs');
+var fs = require('fs-extra');
 var https = require('follow-redirects').https;
 var MemoryStream = require('memorystream');
-var keccak256 = require('js-sha3').keccak256;
+var path = require('path');
+var compilerDir = 'bin';
+
+var yargs = require('yargs')
+  .usage('Usage: [options| version]')
```
cameel: Is this actually supposed to work with a version or only with the whole file name?
Something like `./solcjs-get 0.8.6` or `./solcjs-get 0.8.6` does not not say that the file does not exist. The script tries to download it:
```
Retrieving available version list...
Requested version: v0.8.6
Downloading version v0.8.6
Error downloading file: 404
```

chevdor:
Closing because I don't really have a use for this PR anymore and it is so long ago, I don't even remember the details...

PR #229, opened 2018-06-21
update for travis/macos

tzapu:
hi there
this seems to work ok for me on moth macOS and travis, all tests pass
according to this https://github.com/nodejs/node/issues/19831 in some cases it might work on windows as well, wasn't sure if it's of any concern

the one drawback to this, if you are not pipe ing and expect users to type in stdin, they will need to end with `ctrl+d` the text entry, pipe ing works fine

let me know if this is ok, thank you for all your wonderful work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ node_js:
   - '8'
   - '6'
   - '4'
-branches:
+#branches:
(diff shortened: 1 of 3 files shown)


axic:
Thank you for your patch, this has now been fixed with #127.

PR #238, opened 2018-07-13
Cast contracts to address before accessing address members.

ekpyron:
Dependency of https://github.com/ethereum/solidity/pull/4439.


ekpyron:
Test failure is unrelated to the PR, I think, but I don't have permission to retrigger.

PR #242, opened 2018-07-26
Fix translating linkReferences

yann300:


axic:
@yann300 what was this change for?

axic:
Need to add a test case to verify this. So basically the test case  would use `findLinkReferences` with an old style link reference (`__LibraryName_____...`) ?

yann300:
> So basically the test case would use findLinkReferences with an old style link reference (__LibraryName_____...) 

right I believe so, and the output should be compliant with JSON standard.

PR #245, opened 2018-08-22
Adjust (new) syntax for bare calls

leonardoalt:
Needed in https://github.com/ethereum/solidity/pull/4829


leonardoalt:
Fixed&rebased

PR #246, opened 2018-08-30
Fix --standard-json

y-yamagata:
Fix


westandy-dcp:
I thought the purpose of `--standard-json` was to output a json file from the solcjs compilation.  But this looks like it's tryin to compile JSON provided to the solcjs?

axic:
Thank you for your patch, this has been fixed now by #127.

PR #247, opened 2018-09-05
Support the new solc public methods (solidity_compile/solidity_version/solidity_license)

axic:
Part of https://github.com/ethereum/solidity/issues/3452


axic:
Tested with nightly and 0.4.20 and both work. Since the check for `solidity_compile` is the second and overrides the previous helper, it should always take the latest API (even in binaries supporting both).

PR #256, opened 2018-09-26
Expose the lowlevel compile API

axic:
Part of https://github.com/ethereum/solidity/issues/3452

PR #259, opened 2018-09-26
New breaking api, compile() is an alias to compileStandardWrapper()

axic:
Implements #120 and https://github.com/ethereum/solidity/issues/3452

This is the hard-breaking option, while #122 implements a more incremental change, but carries a lot more baggage.

Depends on #256. Needs test changes.

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -79,20 +79,6 @@ function setupMethods (soljson) {
     };
   }
 
-  var compile = function (input, optimise, readCallback) {
-    var result = '';
-    if (readCallback !== undefined && compileJSONCallback !== null) {
-      result = compileJSONCallback(JSON.stringify(input), optimise, readCallback);
-    } else if (typeof input !== 'string' && compileJSONMulti !== null) {
-      result = compileJSONMulti(JSON.stringify(input), optimise);
-    } else if (compileJSON !== null) {
-      result = compileJSON(input, optimise);
-    } else {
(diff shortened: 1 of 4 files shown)


leonardoalt:
Rebased

axic:
Rebased. This is the last solc-js change for 0.5.0.

chriseth:
I think we should get this merged even without tests because otherwise it stalls another PR on the solidity repo. Also I would say it is fine for tests to temporarily fail on this repo if it allows tests to be continuously green on the solidity repo.

chriseth:
Note that the tests of this PR have to be run against the soljson.js coming out of the still unmerged PR on the solidity side. Locally running only resulted in one test failure related to the package version. Because of that, I will merge this now.

wrapper.js
```diff
@@ -48,25 +48,25 @@ function setupMethods (soljson) {
...
+  if ('x_compileJSONCallback' in soljson) {
     var compileInternal = soljson.cwrap('compileJSONCallback', 'string', ['string', 'number', 'number']);
     compileJSONCallback = function (input, optimize, readCallback) {
       return runWithReadCallback(readCallback, compileInternal, [ input, optimize ]);
     };
   }
 
   var compileStandard = null;
-  if ('_compileStandard' in soljson) {
+  if ('x_compileStandard' in soljson) {
```
axic: This is broken. It was only there for testing to be triggered.

PR #261, opened 2018-09-26
Upgrade dependencies

axic:
Depends on #260.


axic:
It seem fs-extra is not working with Node 4.x.

axic:
Apparently this fixed #135.

Issue #264, opened 2018-09-27
Change circleci config so that coveralls is only run on Node 10


axic:

axic:
@bakaoh do you want to check this one out?

bakaoh:
@axic --- sure, i submitted PR for this issue

PR #265, opened 2018-09-28
Change circleci config so that coveralls is only run on Node 10

bakaoh:
Fix: #264


axic:
This seemingly didn't fix the issue because circleci log show it is run nonetheless, for some reason. Maybe caching?

axic:
Never mind, it seemed to be a caching issue.

PR #266, opened 2018-10-04
Support linking via hash.

chriseth:
Required by https://github.com/ethereum/solidity/pull/5145

### DIFF
--- a/linker.js
+++ b/linker.js
@@ -4,6 +4,10 @@ function keccak256 (input) {
   return keccak('keccak256').update(input).digest();
 }
 
+function libraryHashPlaceholder (input) {
+  return '$' + keccak256(input).toString('hex').slice(0, 34) + '$';
+}
+
 var linkBytecode = function (bytecode, libraries) {
   // NOTE: for backwards compatibility support old compiler which didn't use file names
   var librariesComplete = {};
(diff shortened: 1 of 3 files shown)


linker.js
```diff
@@ -1,3 +1,5 @@
+var sha3 = require('crypto-js/sha3');
```
axic: This should use `ethereumjs-util`, which is already in `package.json`.

chriseth: ethereumjs-util only is a dev dependency, but I think this here needs to be an actual dependency. I want to avoid pulling in all of ethereumjs-util into the dependencies.

axic: Then it should use `keccak.js` (or whatever `ethereumjs-util` is using). `crypto-js/sha3` isn't the best library.

linker.js
```diff
@@ -32,9 +29,19 @@ var linkBytecode = function (bytecode, libraries) {
...
+      // truncate to 37 characters
+      var truncatedName = name.slice(0, 36);
+      var libLabel = '__' + truncatedName + Array(37 - truncatedName.length).join('_') + '__';
+      while (bytecode.indexOf(libLabel) >= 0) {
+        bytecode = bytecode.replace(libLabel, hexAddress);
+      }
+    };
+
+    replace(libraryName);
+    replace(sha3(libraryName, {outputLength: 256}).toString());
```
axic: Should add comments to these two.

test/linker.js
```diff
@@ -82,3 +82,19 @@ tape('Link references', function (t) {
     st.end();
   });
 });
+
+tape('Linker via hash', function (t) {
+  t.test('Non-hashed placeholder', function (st) {
```
axic: This is a duplicated test?

PR #267, opened 2018-10-04
Move determinism test to use compileStandardWrapper

axic:
Part of #120.


leonardoalt:
Please rebase.

axic:
@chriseth please have a look because maybe you know a reason why it was done in that specific way.

chriseth:
The only important aspect is that multiple compilations are done in the same process, and this is still the case, so I guess it's fine.

PR #269, opened 2018-10-04
Do not fail unit tests if network is not available

axic:


test/package.js
```diff
@@ -531,9 +531,14 @@ tape('Compilation', function (t) {
 tape('Loading Legacy Versions', function (t) {
   t.test('loading remote version - development snapshot', function (st) {
     // getting the development snapshot
-    st.plan(3);
+    st.plan(2);
```
leonardoalt: What does `plan` do?

test/package.js
```diff
@@ -531,9 +531,14 @@ tape('Compilation', function (t) {
...
     solc.loadRemoteVersion('latest', function (err, solcSnapshot) {
-      st.notOk(err);
+      if (err) {
+        st.plan(1);
+        st.skip('Network error - skipping remote loading test');
+        st.end();
+        return;
+      }
       var output = solcSnapshot.compile('contract x { function g() public {} }');
```
leonardoalt: Maybe moving `st.plan(2)` here makes it clearer.

axic: It needs to be outside, because it the `loadRemote` is callback based and in case it wouldn't at all fire the callback the test just pass. By having`.plan(2)` outside ensures that is never the case.

PR #270, opened 2018-10-04
Use compileStandardWrapper in solcjs

axic:


axic:
I am pulling the plug, this is fairly risk free. Just opens up the `--standard-json` CLI method to older compiler versions. If an old compiler doesn't support it, it will report a proper error message.

PR #255, opened 2018-09-26
Refactor wrapper (and make compileJSON optional)

axic:
Part of #120 and https://github.com/ethereum/solidity/issues/3452

PR #271, opened 2018-10-04
Support running tests even if compileJSON is missing

axic:
Part of #120. Fixes #255.

### DIFF
--- a/test/package.js
+++ b/test/package.js
@@ -48,6 +48,11 @@ tape('Version and license', function (t) {
 
 tape('Compilation', function (t) {
   t.test('single files can be compiled', function (st) {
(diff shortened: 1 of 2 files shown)


test/package.js
```diff
@@ -48,6 +48,11 @@ tape('Version and license', function (t) {
 
 tape('Compilation', function (t) {
   t.test('single files can be compiled', function (st) {
+    if (!solc.supportsSingle) {
```
leonardoalt: When is this the case?

leonardoalt: Ah, got it now

leonardoalt:
Please rebase

PR #276, opened 2018-10-11
Split prepbulish into prepare and prepublishOnly

chriseth:
@axic I have no idea about the background of this commit, please either merge or close.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   },
   "scripts": {
     "lint": "semistandard",
-    "prepublish": "node downloadCurrentVersion.js && node verifyVersion.js",
+    "prepare": "node downloadCurrentVersion.js && node verifyVersion.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,8 @@
   },
   "scripts": {
     "lint": "semistandard",
-    "prepublish": "node downloadCurrentVersion.js && node verifyVersion.js",
+    "prepare": "node downloadCurrentVersion.js && node verifyVersion.js",
+    "prepublishOnly": "npm run prepare",
```
axic: New node separated the roles and warns when using the old role. However we do need to run it on both cases. I think this PR can wait. Need to figure out the node version this depends on.

axic:
This change breaks node 6 and as it is only a warning on 8/10, it probably is better to leave it till 6 is not supported anymore (probably that's when it will turn into an error).

PR #285, opened 2018-11-08
Use compileStandardWrapper in loadRemoteVersion test

axic:
Part of #120.

### DIFF
--- a/test/package.js
+++ b/test/package.js
@@ -585,15 +585,25 @@ tape('Loading Legacy Versions', function (t) {
         st.end();
         return;
       }
-      if (!solcSnapshot.supportsSingle) {
-        st.plan(1);
-        st.skip('Not supported by solc');
-        st.end();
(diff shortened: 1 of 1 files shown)


test/package.js
```diff
@@ -585,15 +585,25 @@ tape('Loading Legacy Versions', function (t) {
...
+            }
+          }
+        },
+        'sources': {
+          'cont.sol': {
+            'content': 'contract x { function g() public {} }'
+          }
+        }
+      };
+      var output = JSON.parse(solcSnapshot.compileStandardWrapper(JSON.stringify(input)));
```
chriseth: Why not `compile`?

axic: This in preparation for the other change. The big PR which removes this will have a commit turning all these instances into `.compile`.

To me mentally it is easier to follow what's going on as smaller steps.

PR #288, opened 2018-11-08
Show proper error message if wrapper tries compiling multiple files on an old compiler

axic:


axic:
Never mind, this is actually handled earlier.

PR #290, opened 2018-11-08
Add solc.features.* listing

axic:
This is a more consistent way to see if a certain feature supported, than using `supportsXYZ`.

In preparation for #120.

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -230,6 +230,11 @@ function setupMethods (soljson) {
       compileCallback: compileJSONCallback,
       compileStandard: compileStandard
     },
+    features: {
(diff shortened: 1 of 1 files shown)


wrapper.js
```diff
@@ -230,6 +230,11 @@ function setupMethods (soljson) {
       compileCallback: compileJSONCallback,
       compileStandard: compileStandard
     },
+    features: {
+      multipleInputs: compileJSONMulti !== null || compileStandard !== null,
+      importCallback: compileJSONCallback !== null || compileStandard !== null,
+      nativeStandardJSON: compileStandard !== null
```
axic: Any comments on the naming?

chriseth: they are good

PR #293, opened 2018-11-12
Simplify reasons for skipping tests

axic:


axic:
> I don't understand - don't we want to have a check that the feature is supported after the respective version instead of just not running the test if the feature is not present?

After each of those there is the check if the feature is present. So the double check is useless.

axic:
It actually needs two versions, version introduced and version removed. We can go that route too.

Issue #105, opened 2017-04-26
Rewrite solcjs to use functions/callbacks


axic:
And in that case stdio reading can be done properly using:
```
process.stdin.on('data', function(chunk) {
})
process.stdin.on('end', function() {
})
```

PR #303, opened 2018-11-30
Run SMT solver on requested queries

leonardoalt:
and rerun compiler with given responses

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,11 +39,13 @@
   },
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
+    "command-exists": "^1.2.8",
     "fs-extra": "^0.30.0",
     "keccak": "^1.0.2",
     "memorystream": "^0.3.1",
     "require-from-string": "^2.0.0",
     "semver": "^5.5.0",
+    "tmp": "0.0.33",
     "yargs": "^11.0.0"
   },
   "devDependencies": {
--- a/smtchecker.js
+++ b/smtchecker.js
@@ -1,0 +1,65 @@
+var commandExistsSync = require('command-exists').sync;
+var execSync = require('child_process').execSync;
+var fs = require('fs-extra');
+var tmp = require('tmp');
+
+var potentialSolvers = [
+  {
+    name: 'z3',
+    params: ''
+  },
+  {
+    name: 'cvc4',
+    params: '--lang=smt2'
+  }
+];
+var solvers = potentialSolvers.filter(solver => commandExistsSync(solver.name));
+
+function solve (query) {
+  var tmpFile = tmp.fileSync();
+  fs.writeFileSync(tmpFile.name, query);
+  // TODO For now only the first SMT solver found is used.
+  // At some point a computation similar to the one done in
+  // SMTPortfolio::check should be performed, where the results
+  // given by different solvers are compared and an error is
+  // reported if solvers disagree (i.e. SAT vs UNSAT).
+  var solverOutput = execSync(solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name);
+  // Trigger early manual cleanup
+  tmpFile.removeCallback();
+  return solverOutput.toString();
+}
+
+// This function checks the standard JSON output for auxiliaryInputRequested,
+// where smtlib2queries represent the queries created by the SMTChecker.
+// The function runs an SMT solver on each query and adjusts the input for
+// another run.
+// Returns null if no solving is requested.
+function handleSMTQueries (inputJSON, outputJSON) {
+  var auxInputReq = outputJSON.auxiliaryInputRequested;
+  if (!auxInputReq) {
+    return null;
+  }
+
+  var queries = auxInputReq.smtlib2queries;
+  if (!queries || Object.keys(queries) === 0) {
+    return null;
+  }
+
+  if (solvers.length === 0) {
+    throw new Error('No SMT solver available. Assertion checking will not be performed.');
+  }
+
+  var responses = {};
+  for (var query in queries) {
+    responses[query] = solve(queries[query]);
+  }
+
+  // Note: all existing solved queries are replaced.
+  // This assumes that all neccessary queries are quested above.
+  inputJSON.auxiliaryInput = { smtlib2responses: responses };
+  return inputJSON;
+}
+
+module.exports = {
+  handleSMTQueries: handleSMTQueries
+};
--- a/solcjs
+++ b/solcjs
@@ -6,6 +6,7 @@ var originalUncaughtExceptionListeners = process.listeners("uncaughtException");
 var fs = require('fs-extra');
 var path = require('path');
 var solc = require('./index.js');
+var smtchecker = require('./smtchecker.js');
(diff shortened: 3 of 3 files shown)


axic:
There are 3 PRs fixing this and this doesn't work on Node 4 :)

leonardoalt:
What's the workaround then?

axic:
#105 is the solution.

#127 and #246 do the same change as you.

axic:
It seems however that your change is somewhat a 3rd option which indeed works.

axic:
Oh works because ti doesn't have the test! :)

axic:
Can you try adding your version to #127 (just push a commit) ?

leonardoalt:
It should be the same, shouldn't it?

leonardoalt:
Pushed there

solcjs
```diff
@@ -6,6 +6,7 @@ var originalUncaughtExceptionListeners = process.listeners("uncaughtException");
 var fs = require('fs-extra');
 var path = require('path');
 var solc = require('./index.js');
+var SMTChecker = require('./SMTChecker.js');
```
axic: This repo uses lowercase filenames.

SMTChecker.js
```diff
@@ -0,0 +1,56 @@
...
+var fs = require('fs-extra');
+var tmp = require('tmp');
+
+// This function checks the standard JSON output for auxiliaryInputRequested,
+// where smtlib2queries represent the queries created by the SMTChecker.
+// The function runs an SMT solver on each query and compiles the input
+// again providing the answers given by the solvers.
+function handleSMTQueries(input, output) {
+  var outputJSON = JSON.parse(output);
+  var auxInputReq = outputJSON["auxiliaryInputRequested"];
```
axic: Can just use `outputJSON.auxiliarInputRequested` here and below.

SMTChecker.js
```diff
@@ -0,0 +1,56 @@
...
+  // reported if solvers disagree (i.e. SAT vs UNSAT).
+  var potentialSolvers = ["z3", "cvc4"];
+  var solvers = [];
+  for (var solver in potentialSolvers)
+    if (commandExistsSync(potentialSolvers[solver])) {
+      solvers.push(potentialSolvers[solver]);
+      break;
+    }
+  if (solvers.length == 0) {
+    console.log("No SMT solver available. Assertion checking will not be performed.");
```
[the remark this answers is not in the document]
leonardoalt: Which function should I use?

axic:
This is my proposal:
```
// Returns `null` if nothing is to be solved or a new `input` if it needs to be run again.
function handleSMTQueries(input, output, solver) {
  if (!output.auxiliaryInputRequested)
    return null

  // Could also just throw an exception here?
  if (!solverCallback)
    return null

  let queries = output.auxiliaryInputRequested.smtlib2queries
  var responses = {}
  for (var queryKey in queries) {
    responses[queryKey] = solver(queries[queryKey])
  }

  input.auxiliaryInput.smtlib2esponses = responses
  return input
}
```

SMTChecker.js
```diff
@@ -0,0 +1,56 @@
...
+
+  var queries = outputJSON["auxiliaryInputRequested"]["smtlib2queries"];
+  var responses = {};
+  if (!queries || Object.keys(queries) == 0)
+	return output;
+  else {
+    for (var query in queries) {
+      var tmpFile = tmp.fileSync();
+      fs.writeFileSync(tmpFile.name, queries[query]);
+      var solverOutput = execSync(solvers[0] + ' ' + tmpFile.name);
```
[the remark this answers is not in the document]
leonardoalt: Why callback? Is it ok to have everything sync?

leonardoalt:
`input.auxiliaryInput.smtlib2esponses = responses`
doesn't work because `input.auxiliaryInput` is still `undefined` at that point

axic:
It is the high level design :)

leonardoalt:
Updated extracting the `solve` part into a sync function

leonardoalt:
not sure about the changes to `solcjs`, it's gotten quite long with the error handling

smtchecker.js
```diff
@@ -0,0 +1,57 @@
...
+// The function runs an SMT solver on each query and adjusts the input for
+// another run.
+// Returns null if no solving is requested.
+function handleSMTQueries (inputJSON, outputJSON) {
+  var auxInputReq = outputJSON.auxiliaryInputRequested;
+  if (!auxInputReq)
+    return null;
+
+  if (solvers.length == 0)
+    throw "No SMT solver available. Assertion checking will not be performed."
```
axic: What syntax is this?

solcjs
```diff
@@ -52,7 +53,27 @@ function abort (msg) {
 
 if (argv['standard-json']) {
   var input = fs.readFileSync(process.stdin.fd);
-  console.log(solc.compile(input.toString('utf8')))
+  var output = solc.compileStandardWrapper(input.toString('utf8'));
+
+  try {
+    var inputJSON = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output));
+    if (inputJSON)
+      output = solc.compileStandardWrapper(JSON.stringify(inputJSON));
```
axic: This should not be wrapped in this try-catch.

solcjs
```diff
@@ -52,7 +53,27 @@ function abort (msg) {
 
 if (argv['standard-json']) {
   var input = fs.readFileSync(process.stdin.fd);
-  console.log(solc.compile(input.toString('utf8')))
+  var output = solc.compileStandardWrapper(input.toString('utf8'));
+
+  try {
+    var inputJSON = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output));
```
axic: Could just use `input` here (to override the one above)

package.json
```diff
@@ -39,11 +39,13 @@
   },
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
+    "command-exists": "^1.2.8",
     "fs-extra": "^0.30.0",
     "keccak": "^1.0.2",
     "memorystream": "^0.3.1",
     "require-from-string": "^2.0.0",
     "semver": "^5.5.0",
+    "tmp": "0.0.33",
```
axic: Based on this version number this seems to be a non-maintained dependency. We have to be careful what kind of dependencies we take.

smtchecker.js
axic: `throw new Exception("...")`

But better to use `assert("No SMT..")`

package.json
leonardoalt: Do you know any other lib for this? When I searched it seemed the most common

smtchecker.js
leonardoalt: but we don't want it to crash, the queries are simply not solved then

axic: Assert is just a basic exception.

package.json
axic: Ok it has a fair amount of users.

axic:
Reordered some things and fixed lint.

@leonardoalt can you double check it works?

Can you create an issue for adding tests? That will need a manual installation of cvc4/z3.

leonardoalt:
Yes, it works!

PR #312, opened 2018-12-04
Link bytecode using references

### LINKED ISSUE #193: Update linkBytecode to support linkReferences structure from jsonio

jhokkar:
Support to provide `linkReferences` as optional third parameter for `linkBytecode`. If it is not provided then `findLinkReferences` is used to find it. If library name is not present in `linkReferences` then it falls back to manual search.

Fixes  #193


linker.ts
```diff
@@ -55,10 +65,12 @@ function replacePlaceholder (bytecode, label, address) {
...
  *
+ * @param linkReferences A optional mapping of libraries to lists of placeholder positions in the binary.
+ *
  * @returns bytecode Hex-encoded bytecode string with placeholders replaced with addresses.
  *    Note that some placeholders may remain in the bytecode if `libraries` does not provide addresses for all of them.
  */
-function linkBytecode (bytecode: string, libraries: LibraryAddresses): string {
+function linkBytecode (bytecode: string, libraries: LibraryAddresses, linkReferences?: LinkReferences): string {
   assert(typeof bytecode === 'string');
   assert(typeof libraries === 'object');
```
matheusaaguiar, line 76: Since other param types are being asserted, should we do the same with `linkReferences`?
```
assert(!linkReferences || typeof linkReferences === 'object');
```
Not sure if that makes sense in this case...

cameel:
@r0qs I see there's a merge commit. Can you rebase it on top of `master` instead?

linker.ts
```diff
@@ -27,14 +27,24 @@ function libraryHashPlaceholder (fullyQualifiedLibraryName) {
...
+    linkReferences = findLinkReferences(bytecode);
+  }
+
   // truncate to 36 characters
   const truncatedName = label.slice(0, 36);
-  const libLabel = `__${truncatedName.padEnd(36, '_')}__`;
 
-  while (bytecode.indexOf(libLabel) >= 0) {
-    bytecode = bytecode.replace(libLabel, address);
+  if (linkReferences && linkReferences[truncatedName]) {
```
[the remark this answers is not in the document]
r0qs: Indeed! Thanks.

linker.ts
r0qs: I think this is unnecessary here for the `linkReferences`, since it is an optional parameter, and it is only passed to `replacePlaceholder` where its existence is checked.

r0qs:
> @r0qs I see there's a merge commit. Can you rebase it on top of `master` instead?

Yes. I did it now, but as the commits were very old and the file structures changed significantly, the OP changes needed to be discarded during the rebase. However, I added the OP as a co-author of my changes, since I reused part of his solution.

matheusaaguiar (approved):
lgtm

PR #316, opened 2018-12-07
Provide an SMT solver callback to solc

leonardoalt:
Maybe this is what was intended in the first place.
If the smtchecker is used only by `solcjs --standard-json` the development frameworks still don't use it. By moving it to `compileStandardWrapper` it works out of the box (when it's released and the frameworks update their solc versions).
Example using truffle (after updating solc manually):
```
$ truffle compile
Compiling ./contracts/assert.sol...

Compilation warnings encountered:

Warning: This is a pre-release compiler version, please do not use it in production.
,/home/leonardo/devel/truffle_test/contracts/assert.sol:7:3: Warning: Assertion violation happens here
		assert(x > 0);
		^-----------^
  for:
  x = 0
```

### DIFF
--- a/smtchecker.js
+++ b/smtchecker.js
@@ -26,7 +26,7 @@ function solve (query) {
   var solverOutput = execSync(solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name);
   // Trigger early manual cleanup
   tmpFile.removeCallback();
-  return solverOutput.toString();
+  return { contents: outputStr };
 }
 
 // This function checks the standard JSON output for auxiliaryInputRequested,
@@ -51,7 +51,7 @@ function handleSMTQueries (inputJSON, outputJSON) {
 
   var responses = {};
   for (var query in queries) {
-    responses[query] = solve(queries[query]);
+    responses[query] = solve(queries[query]).contents;
   }
 
   // Note: all existing solved queries are replaced.
@@ -61,5 +61,6 @@ function handleSMTQueries (inputJSON, outputJSON) {
 }
 
 module.exports = {
-  handleSMTQueries: handleSMTQueries
+  handleSMTQueries: handleSMTQueries,
+  smtSolve: solve
 };
--- a/solcjs
+++ b/solcjs
@@ -5,6 +5,7 @@ var originalUncaughtExceptionListeners = process.listeners("uncaughtException");
 
 var fs = require('fs-extra');
 var path = require('path');
+
 var solc = require('./index.js');
 var smtchecker = require('./smtchecker.js');
 // FIXME: remove annoying exception catcher of Emscripten
@@ -54,29 +55,26 @@ function abort (msg) {
 if (argv['standard-json']) {
   var input = fs.readFileSync(process.stdin.fd).toString('utf8');
   var output = solc.compileStandardWrapper(input);
-
   try {
-    var inputJSON = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output));
-    if (inputJSON) {
-      output = solc.compileStandardWrapper(JSON.stringify(inputJSON));
+    input = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output));
+    if (input !== null) {
+      output = solc.compileStandardWrapper(JSON.stringify(input));
     }
-  }
-  catch (e) {
+  } catch (e) {
     var addError = {
-      component: "general",
+      component: 'general',
       formattedMessage: e.toString(),
       message: e.toString(),
-      type: "Warning"
+      type: 'Warning'
     };
 
-    var outputJSON = JSON.parse(output);
-    if (!outputJSON.errors) {
-      outputJSON.errors = []
+    output = JSON.parse(output);
+    if (!output.errors) {
+      output.errors = [];
     }
-    outputJSON.errors.push(addError);
-    output = JSON.stringify(outputJSON);
+    output.errors.push(addError);
+    output = JSON.stringify(output);
   }
-
   console.log(output);
   process.exit(0);
 } else if (files.length === 0) {
@@ -98,20 +96,24 @@ for (var i = 0; i < files.length; i++) {
   }
 }
 
-var output = JSON.parse(solc.compileStandardWrapper(JSON.stringify({
-  language: 'Solidity',
-  settings: {
-    optimizer: {
-      enabled: argv.optimize
-    },
-    outputSelection: {
-      '*': {
-        '*': [ 'abi', 'evm.bytecode' ]
+var output = JSON.parse(solc.compileStandardWrapper(
+  JSON.stringify({
+    language: 'Solidity',
+    settings: {
+      optimizer: {
+        enabled: argv.optimize
+      },
+      outputSelection: {
+        '*': {
+          '*': [ 'abi', 'evm.bytecode' ]
+        }
       }
-    }
-  },
-  sources: sources
-})));
+    },
+    sources: sources
+  }), {
+    smtCallback: smtchecker.smtSolve
+  }
+));
 
 if (!output) {
   abort('No output from compiler');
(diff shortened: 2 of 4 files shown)


axic:
I don't think this is a good idea, because the wrapper has no business in executing filesystem stuff. What I meant by having a callback for the `solver` is that callback could be provided in the `compileStandardWrapper`.

leonardoalt:
But who registeres the callback? Frameworks use `compileStandardWrapper` directly, so where would `handleSMTQueries` be called?

leonardoalt:
Or do you mean for `compileStandardWrapper` to register the callback?

axic:
> But who registeres the callback?

It is supplied to `compileStandardWrapper`. The node file access tools cannot be compiled for a browser environment and that means the current version of the PR breaks on Remix for example.

leonardoalt:
Ok, I understand now, makes sense.
But then again, who would supply the "smt callback" to `compileStandardWrapper`? Truffle, eg?

axic:
My new API proposal is: `compiler(standardJson, callbacks)` where `callbacks` is an object having two optional fields (`importCallback`, `smtSolverCallback`).

If the `callbacks` argument is not an object, but a function, then it is considered as `importCallback`.

Then truffle for example would have the opportunity to call `compile` with the smt solver:
```
var solc = require('solc')
var smtSolver = require('solc/smtSolver')

var output = solc.compile(input, { importCallback: resolveImports, smtSolverCallback: smtSolver.execResolver })
```

It has to be driven by the caller.

leonardoalt:
@axic I changed this PR to the new API you proposed. Moreover, on the Solidity side, added a new callback to be used by `SmtLib2Interface` as a sync SMT solver (https://github.com/ethereum/solidity/pull/5642), to solve everything in one run. If the callback is not given, it still uses the JSON queries with the double run (if standard JSON).
In this PR `solcjs` gives `solc` the `smtCallback` by default.
Frameworks can call the compiler in a similar way if they run locally.

leonardoalt:
This PR should only be merged iff https://github.com/ethereum/solidity/pull/5642 is merged.

axic:
Why is the Solidity change required? My proposal didn't require that change :)

leonardoalt:
I know, I talked with @chriseth and @ekpyron and we decided to give the single run a go, so I extended your proposal for that :p

axic:
We had this double callback system initially (months ago) and decided it wasn't the best.

solcjs
```diff
@@ -54,29 +55,26 @@ function abort (msg) {
...
+  } catch (e) {
     var addError = {
-      component: "general",
-      formattedMessage: e.toString(),
-      message: e.toString(),
-      type: "Warning"
+      'component': 'general',
+      'formattedMessage': e.toString(),
+      'message': e.toString(),
+      'type': 'Warning'
```
axic: Keys don't need to be quoted in javascript.

wrapper.js
```diff
@@ -40,54 +51,65 @@ function setupMethods (soljson) {
...
+    if (cbSmt !== undefined) {
+      removeFunction(cbSmt);
+    }
     return output;
   };
 
   var compileJSON = null;
   if ('_compileJSON' in soljson) {
-    compileJSON = soljson.cwrap('compileJSON', 'string', ['string', 'number']);
+    compileJSON = soljson.cwrap('compileJSON', 'string', ['string', 'number', 'number']);
```
axic: These old APIs cannot be changed. It breaks backwards compatibility.

wrapper.js
```diff
@@ -40,54 +51,65 @@ function setupMethods (soljson) {
...
   if ('_compileStandard' in soljson) {
-    var compileStandardInternal = soljson.cwrap('compileStandard', 'string', ['string', 'number']);
+    var compileStandardInternal = soljson.cwrap('compileStandard', 'string', ['string', 'number', 'number']);
     compileStandard = function (input, readCallback) {
       return runWithReadCallback(readCallback, compileStandardInternal, [ input ]);
     };
   }
   if ('_solidity_compile' in soljson) {
-    var solidityCompile = soljson.cwrap('solidity_compile', 'string', ['string', 'number']);
+    var solidityCompile = soljson.cwrap('solidity_compile', 'string', ['string', 'number', 'number']);
```
axic: We cannot just change this. We need to define a new function name. This breaks backwards compatibility. There doesn't seem to be a way to detect the signature of a given symbol. Well we could detect the signature based on what `solidity_version` returns.

leonardoalt:
Ok, before we go on with reviewing, we should decide then if want this or not.

axic:
This new version of the PR doesn't actually support what I proposed. There is no way to use local solver as a callback.

leonardoalt:
I don't understand why not. It's working fine with my local runs. Of course it won't work on browsers, but with frameworks that run locally it should, shouldn't it?

leonardoalt:
https://github.com/ethereum/solidity/pull/5675

smtchecker.js
```diff
@@ -6,7 +6,7 @@ var tmp = require('tmp');
 var potentialSolvers = [
   {
     name: 'z3',
-    params: ''
+    params: '-smt2'
```
axic: Is this a separate fix?

smtchecker.js
```diff
@@ -16,17 +16,37 @@ var potentialSolvers = [
 var solvers = potentialSolvers.filter(solver => commandExistsSync(solver.name));
 
 function solve (query) {
-  var tmpFile = tmp.fileSync();
+  var tmpFile = tmp.fileSync({ postfix: '.smt2' });
```
axic: Is this a separate fix?

smtchecker.js
```diff
@@ -16,17 +16,37 @@ var potentialSolvers = [
...
   // At some point a computation similar to the one done in
   // SMTPortfolio::check should be performed, where the results
   // given by different solvers are compared and an error is
   // reported if solvers disagree (i.e. SAT vs UNSAT).
-  var solverOutput = execSync(solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name);
+  var outputStr;
+  try {
+    var solverOutput = execSync(
+      solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name, {
+        timeout: 10000
```
axic: Is this a separate fix?

smtchecker.js
```diff
@@ -16,17 +16,37 @@ var potentialSolvers = [
...
+    // execSync throws if the process times out or returns != 0.
+    // The latter might happen with z3 if the query asks for a model
+    // for an UNSAT formula. We can still use stdout.
+    outputStr = e.stdout.toString();
+    if (
+      !outputStr.startsWith('sat') &&
+      !outputStr.startsWith('unsat') &&
+      !outputStr.startsWith('unknown')
+    ) {
+      throw new Error('Failed solve SMT query. ' + e.toString());
```
axic: Is this a separate fix?

leonardoalt: This yes, the others are rather enhancements.

axic: Can you create a PR for these fixes? Perhaps on top of #325.

leonardoalt: https://github.com/ethereum/solc-js/pull/328

axic:
Removed the "z3 timeout" change from this. It still needs to be rebased if we want to continue with this.

axic:
@leonardoalt can this be rebased? is it worth rebasing?

leonardoalt:
We can close it.

elenadimitrova:
Ability to easily compile contracts with smt solver is a major barrier to entry. I don't understand why this is being closed is there a replacement issue to track this gap?

chriseth:
@elenadimitrova https://github.com/ethereum/solidity/pull/5675

chriseth:
@elenadimitrova perhaps to clarify: This is a change in the API and thus we need to wait for the next breaking release.

leonardoalt:
I'm working on this again.
https://github.com/ethereum/solidity/pull/7447

leonardoalt:
Too old to rebase.
Closing in favor of https://github.com/ethereum/solc-js/pull/382

PR #319, opened 2018-12-12
Add SMTChecker smoke tests

axic:
Part of #307.

This also ensures that code coverage includes `smtchecker.js`


axic:
This is harmless, merging.

PR #323, opened 2018-12-18
Fix translate.escapeString to actually replace all occurrances

axic:
Actually fix #318.

### DIFF
--- a/translate.js
+++ b/translate.js
@@ -150,9 +150,9 @@ function translateJsonCompilerOutput (output, libraries) {
 
 function escapeString (text) {
   return text
-    .replace(/\n/, '\\n')
-    .replace(/\r/, '\\r')
-    .replace(/\t/, '\\t');
+    .replace(/\n/g, '\\n')
+    .replace(/\r/g, '\\r')
+    .replace(/\t/g, '\\t');
 }
 
 function formatAssemblyText (asm, prefix, source) {


test/fixtureAsmJson.output
```diff
@@ -1,104 +1,104 @@
...
+      PUSH 0			ne her
+      ADDRESS 			\n\n\n 
+      PUSH FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF			\n\n\n  f
+      AND 			\n\n\n  f
+      PUSH 26121FF0			\n\n\n  f
+      SWAP1 			\n\n\n  function f
+      POP 			\n\n\n  function f
+      PUSH 100000000000000000000000000000000000000000000000000000000			\n\n\n  function f
+      MUL 			\n\n\n  function f
+      SWAP1 			ionally\n\n\n  function f
```
chriseth: This looks wrong

test/fixtureAsmJson.sol
```diff
@@ -1,4 +1,7 @@
 contract C {
+  // Leaving an empty line here intentionally
+
+
   function f() returns (bytes4) {
```
chriseth: no public?

axic: This is not compiled, it is just for formatting.

chriseth:
Could you change this so that the comment is moved further down? Then the test would be proper already for the first `PUSH1 0x40`.

test/fixtureAsmJson.output
chriseth: It was correct before. It might be that `\n` is counted as two bytes?

test/fixtureAsmJson.sol
chriseth: I don't understand, there is assembly above. I thought it would be the test expectation.

axic: It is not compiled by the compiler, only used by the formatting tool.

chriseth:
I currently cannot verify that the test actually checks the change.

axic:
The reason this is wrong that forgot to update `fixtureAsmJson.json`.

PR #328, opened 2018-12-20
Fix SMT when Z3 returns UNSAT

leonardoalt:


smtsolver.js
```diff
@@ -20,17 +20,37 @@ function solve (query) {
...
+      solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name, {
+        timeout: 10000
+      }
+    );
+    outputStr = solverOutput.toString();
+  } catch (e) {
+    // execSync throws if the process times out or returns != 0.
+    // The latter might happen with z3 if the query asks for a model
+    // for an UNSAT formula. We can still use stdout.
+    outputStr = e.stdout.toString();
```
axic: Does this contain a proper output?

smtsolver.js
```diff
@@ -20,17 +20,37 @@ function solve (query) {
...
-  var tmpFile = tmp.fileSync();
+  var tmpFile = tmp.fileSync({ postfix: '.smt2' });
   fs.writeFileSync(tmpFile.name, query);
   // TODO For now only the first SMT solver found is used.
   // At some point a computation similar to the one done in
   // SMTPortfolio::check should be performed, where the results
   // given by different solvers are compared and an error is
   // reported if solvers disagree (i.e. SAT vs UNSAT).
-  var solverOutput = execSync(solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name);
+  var outputStr;
```
axic: Would just name this `output` or `ret`.

smtsolver.js
leonardoalt: It might, that's why we check

PR #331, opened 2019-01-14
Remove soljson.js if download is interrupted

axic:


axic:
Never mind this was already merge, this was a bad rebase when I tried to clean up my local repos.

PR #339, opened 2019-02-11
Force tape version to 4.9.2.

chriseth:
A new tape version was released two days ago which makes our test setup fail. I'm still not sure whether it is a bug with our tests or with tape, but this should get us through the release tomorrow.

I'm suspecting [this PR](https://github.com/substack/tape/pull/403) being the cause of it.


chriseth:
Created: https://github.com/ethereum/solidity/issues/5963

PR #353, opened 2019-04-29
Version 0.4.26

chriseth:


blitz-1306:
Hello, @chriseth. Just curious: this would be merged any time soon?

chriseth:
Yeah good question. It has been superseded by more recent versions. As far as I know, npm only records the latest version. It could be merged in a way where the merge commit basically undoes the change here, so that the commit is at least an ancestor of HEAD... Do you have any other suggestions?

blitz-1306:
Have you considered using https://stackoverflow.com/a/27665031/2269749 ?

chriseth:
Ah, so you are saying the problem is that the version is not published on NPM (and not that this branch is not merged)?

chriseth:
Thanks for the hint, just published it!

blitz-1306:
@chriseth You mentioned:
> As far as I know, npm only records the latest version.
That is why I suggested to consider tags. Also, I'm interested in Solc release of 0.4.26, which is not currently present in SolcJs. One of my projects allows to run solidity legacy versions.

You're welcome.

chriseth:
So is this problem fixed for you?

blitz-1306:
@chriseth yeah, sure.

chriseth:
Ok, great!

axic:
@chriseth I usually push the tags to this repo, but it hasn't been done since February.

Did you push this PR as the old version? If so, we should just tag this commit hash 0.4.26 and close the PR.

chriseth:
Yes, the npm packages was published from this commit. Can you push the tags, please? Also, could you update the solidity release checklist to include that step? Thanks!

axic:
Pushed all the missing tags.

chriseth:
Thanks!

PR #363, opened 2019-05-28
Set version to 0.5.9.

chriseth:


chriseth:
Trouble with CircleCI. Releasing with confirmed local tests only.

PR #367, opened 2019-06-10
fix: ensure CLI replaces path separator on Windows platforms correctly

### LINKED ISSUE #365: Solcjs can't write output on Windows machines due to path separators
Looks like there's a bug in Solcjs with regards to how it handles the file names of output files generated by solc.

Here's what's happening:

Assuming we run `solcjs` cli like this on Linux/Mac:

```
$ solcjs --abi --bin -o some-folder path/to/file.sol
```

It'll generate:

```
some-folder/path_to_file_ContractName.abi
some-folder/path_to_file_ContractName.bin
```
Notice how slashes are replaced with underscores.

On windows however, often backslashes are used so the same command looks like this:

```
$ solcjs --abi --bin -o some-folder path\to\file.sol
```

However, writing the output fails because it doesn't replace backslashes with underscores. This is what it tries to write:

```
some-folder\path\to\file_ContractName.abi
some-folder\path\to\file_ContractName.bin
```

This won't work because `path\fo\` doesn't exist inside `some-folder`.

I tested this with `solc` (instead of `solcjs`) and can confirm that it behaves correctly on Windows (replaces backslashes as expected).

I'm happy to look into this if this is fixable. I thought solcjs is only wrapping solc but maybe it does more to it.

Any ideas, pointers?

PascalPrecht:
Fixes #365

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -141,7 +141,7 @@ function writeFile (file, content) {
 for (var fileName in output.contracts) {
   for (var contractName in output.contracts[fileName]) {
     var contractFileName = fileName + ':' + contractName;
-    contractFileName = contractFileName.replace(/[:./]/g, '_');
+    contractFileName = contractFileName.replace(/[:./\\]/g, '_');
 
     if (argv.bin) {
       writeFile(contractFileName + '.bin', output.contracts[fileName][contractName].evm.bytecode.object);


chriseth:
Thanks!

PascalPrecht:
Wonderful, thanks for merging this @chriseth !

Do you know when a release can be expected, that includes this fix?

chriseth:
The releases are synchronized with Solidity's releases. We haven't set a release date yet, but should be in about 2 weeks.

PascalPrecht:
Hey @chriseth 

just wanted to check in, any updates on the release? :)
If you think this won't happen any time soon, that's fine too. Will have to reference an exact commit that for CI.

chriseth:
@PascalPrecht we released on npm last week, it should contain this PR.

PascalPrecht:
Ah, thank you @chriseth ! 🙏 I didn't even check haha

PR #369, opened 2019-07-04
Test on node.js v12 too

axic:


axic:
This is successful and should be harmless, merging.

axic:
> What's the reason why only even version numbers are used?

Even numbers are the stable releases. Odd numbers are the unstable releases.

erak:
Ah, thanks!

PR #382, opened 2019-09-17
Add callback kind to solc

leonardoalt:
Depends on #410

New version of https://github.com/ethereum/solc-js/pull/316

Depends on https://github.com/ethereum/solidity/pull/7447

With this change, one can call
`compileStandard(..., smtchecker.smtCallback(smtsolver.smtSolver), ...)`
and get the SMTChecker working only requiring a local `z3` or `cvc4`.

### DIFF
--- a/smtsolver.js
+++ b/smtsolver.js
@@ -46,7 +46,7 @@ function solve (query) {
       !solverOutput.startsWith('unsat') &&
       !solverOutput.startsWith('unknown')
     ) {
-      throw new Error('Failed solve SMT query. ' + e.toString());
+      throw new Error('Failed to solve SMT query. ' + e.toString());
     }
   }
   // Trigger early manual cleanup
@@ -55,5 +55,6 @@ function solve (query) {
 }
 
 module.exports = {
-  smtSolver: solve
+  smtSolver: solve,
+  availableSolvers: solvers.length
 };
--- a/test/smtcallback.js
+++ b/test/smtcallback.js
@@ -1,0 +1,165 @@
+const tape = require('tape');
+const fs = require('fs');
+const path = require('path');
+const semver = require('semver');
+const solc = require('../index.js');
+const smtchecker = require('../smtchecker.js');
+const smtsolver = require('../smtsolver.js');
+
+var pragmaSol = 'pragma solidity >=0.0;\n';
+var pragmaSMT = 'pragma experimental SMTChecker;\n';
+
+function collectErrors (solOutput) {
+  if (solOutput === undefined) {
+    return [];
+  }
+
+  var errors = [];
+  for (var i in solOutput.errors) {
+    var error = solOutput.errors[i];
+    if (error.message.includes('This is a pre-release compiler version')) {
+      continue;
+    }
+    errors.push(error.message);
+  }
+  return errors;
+}
+
+function expectErrors (errors, expectations) {
+  if (errors.length !== expectations.length) {
+    return false;
+  }
+
+  for (var i in errors) {
+    if (!errors[i].includes(expectations[i])) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+tape('SMTCheckerCallback', function (t) {
+  t.test('Interface via callback', function (st) {
+    if (!semver.gt(solc.semver(), '0.5.99')) {
+      st.skip('SMT callback not implemented by this compiler version.');
+      st.end();
+      return;
+    }
+
+    var satCallback = function (query) {
+      return { contents: 'sat\n' };
+    };
+    var unsatCallback = function (query) {
+      return { contents: 'unsat\n' };
+    };
+    var errorCallback = function (query) {
+      return { error: 'Fake SMT solver error.' };
+    };
+
+    var input = { 'a': { content: pragmaSol + pragmaSMT + 'contract C { function f(uint x) public pure { assert(x > 0); } }' } };
+    var inputJSON = JSON.stringify({
+      language: 'Solidity',
+      sources: input
+    });
+    var tests = [
+      { cb: satCallback, expectations: ['Assertion violation happens here'] },
+      { cb: unsatCallback, expectations: [] },
+      { cb: errorCallback, expectations: ['BMC analysis was not possible'] }
+    ];
+    for (var i in tests) {
+      var test = tests[i];
+      var output = JSON.parse(solc.compile(
+        inputJSON,
+        { smtSolver: test.cb }
+      ));
+      var errors = collectErrors(output);
+      st.ok(expectErrors(errors, test.expectations));
+    }
+    st.end();
+  });
+
+  t.test('Solidity smtCheckerTests', function (st) {
+    var testdir = path.resolve(__dirname, 'smtCheckerTests/');
+    if (!fs.existsSync(testdir)) {
+      st.skip('SMT checker tests not present.');
+      st.end();
+      return;
+    }
+
+    if (smtsolver.availableSolvers === 0) {
+      st.skip('No SMT solver available.');
+      st.end();
+      return;
+    }
+
+    var sources = [];
+
+    // BFS to get all test files
+    var dirs = [testdir];
+    var i;
+    while (dirs.length > 0) {
+      var dir = dirs.shift();
+      var files = fs.readdirSync(dir);
+      for (i in files) {
+        var file = path.join(dir, files[i]);
+        if (fs.statSync(file).isDirectory()) {
+          dirs.push(file);
+        } else {
+          sources.push(file);
+        }
+      }
+    }
+
+    // Read tests and collect expectations
+    var tests = [];
+    for (i in sources) {
+      st.comment('Collecting ' + sources[i] + '...');
+      var source = fs.readFileSync(sources[i], 'utf8');
+      var expected = [];
+      var delimiter = '// ----';
+      if (source.includes(delimiter)) {
+        expected = source.substring(source.indexOf('// ----') + 8, source.length).split('\n');
+        // Sometimes the last expectation line ends with a '\n'
+        if (expected.length > 0 && expected[expected.length - 1] === '') {
+          expected.pop();
+        }
+      }
+      tests[sources[i]] = {
+        expectations: expected,
+        solidity: { test: { content: pragmaSol + source } }
+      };
+    }
+
+    // Run all tests
+    for (i in tests) {
+      var test = tests[i];
+      var output = JSON.parse(solc.compile(
+        JSON.stringify({
+          language: 'Solidity',
+          sources: test.solidity
+        }),
+        { smtSolver: smtchecker.smtCallback(smtsolver.smtSolver) }
+      ));
+      st.ok(output);
+
+      // Collect obtained error messages
+      test.errors = collectErrors(output);
+
+      // These are errors in the SMTLib2Interface encoding.
+      if (test.errors.length > 0 && test.errors[test.errors.length - 1].includes('BMC analysis was not possible')) {
+        continue;
+      }
+
+      // These are due to CHC not being supported via SMTLib2Interface yet.
+      if (test.expectations.length !== test.errors.length) {
+        continue;
+      }
+
+      // Compare expected vs obtained errors
+      st.ok(expectErrors(test.expectations, test.errors));
+    }
+
+    st.end();
+  });
+});
--- a/wrapper.js
+++ b/wrapper.js
@@ -3,6 +3,7 @@ var translate = require('./translate.js');
 var requireFromString = require('require-from-string');
 var https = require('https');
 var MemoryStream = require('memorystream');
+var semver = require('semver');
 
 function setupMethods (soljson) {
   var version;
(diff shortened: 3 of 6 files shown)


leonardoalt:
Still needs tests.

wrapper.js
```diff
@@ -26,7 +26,7 @@ function setupMethods (soljson) {
   };
 
   // This calls compile() with args || cb
-  var runWithReadCallback = function (readCallback, compile, args) {
+  var runWithReadCallback = function (readCallback, auxInputCallback, compile, args) {
```
leonardoalt: @axic new argument here or at the end?

leonardoalt:
Do we want to give this callback by default if none is provided?

leonardoalt:
Current state:
If a function is given to `runWithCallbacks` it uses it as the old `readCallback`, so nothing changes there.
If an object `callbacks` is given to `runWithCallbacks`, use `callbacks.importCallback` as the old `readCallback` and `callbacks.smtSolverCallback` as the new SMT callback.

In any case, they are wrapped into a single callback that decodes the required callback type. This is the breaking change in the data. Solidity will send the callback type before the data:
- import callback `sources:`
- SMT solver callback `smt-query:`

leonardoalt:
@axic @chriseth

leonardoalt:
The tests added here require
- `z3` or `cvc4` installed locally
- `smtCheckerTests` in `test`
Do we want the `solc-js` tests do take care of these, or should the Solidity solc-js tests do that?

leonardoalt:
Tentative decisions from call:
- Instead of encoding the callback kind as a prefix in the data, break the C interface and add a new parameter for the callback kind.
- The above implies we need to check the version to decide the number of arguments for `solidity_compile` from `soljson`.
- Break the solc-js `compileStandardWrapper`: only take an object of callbacks instead of `function || object`.
- solc-js runs simple SMT tests just to test interface, can install solver in CI just for tests
- Solidity runs the `smtCheckerTests` with `SMTLib2Interface` by copying `smtCheckerTests` to `solc-js/test/` and running the solc-js tests.

leonardoalt:
The tests here will only pass with `soljson.js` from https://github.com/ethereum/solidity/pull/7447

wrapper.js
```diff
@@ -14,8 +14,8 @@ function setupMethods (soljson) {
 
   var wrapCallback = function (callback) {
     assert(typeof callback === 'function', 'Invalid callback specified.');
-    return function (path, contents, error) {
-      var result = callback(soljson.Pointer_stringify(path));
+    return function (kind, data, contents, error) {
+      var result = callback(soljson.Pointer_stringify(kind), soljson.Pointer_stringify(data));
```
axic: Will we add the context too? In a separate PR?

smtsolver.js
```diff
@@ -53,5 +53,6 @@ function solve (query) {
 }
 
 module.exports = {
-  smtSolver: solve
+  smtSolver: solve,
+  availableSolvers: solvers
```
axic: Why is this needed? And if it is needed, would it make sense mapping it into names only?

wrapper.js
```diff
@@ -26,20 +26,48 @@ function setupMethods (soljson) {
   };
 
   // This calls compile() with args || cb
-  var runWithReadCallback = function (readCallback, compile, args) {
+  var runWithCallbacks = function (callbacks, compile, args) {
+    var readCallback;
+    var smtSolverCallback;
+
+    if (typeof callbacks === 'function') {
```
axic: Did we decide to not have this workaround in 0.6 then because we allow people to upgrade to the new syntax in 0.5?

wrapper.js
leonardoalt: Yeah I guess separate PR is better

smtsolver.js
leonardoalt: It is needed by the `smtcallback.js` test such that it can decide not to run the tests if there's no solver installed. Names only should be fine. Even just the amount would be fine

wrapper.js
leonardoalt: Yea need to rebase and will remove this

smtsolver.js
leonardoalt: Updated

leonardoalt:
@axic This is ready for review, but the test that fails is not going to work without `soljson.js` from Solidity `develop_060`

test/smtcallback.js
```diff
@@ -0,0 +1,156 @@
...
+          sources: test.solidity
+        }),
+        { smtSolverCallback: smtchecker.smtCallback(smtsolver.smtSolver) }
+      ));
+      st.ok(output);
+
+      // Collect obtained error messages
+      test.errors = collectErrors(output);
+
+      // Compare expected vs obtained errors
```
erak: Is this supposed to be in line 150?

axic:
Why is this still done on `master_060` (and/or why is `master_060` out of date compared to `master`)?

leonardoalt:
Aren't we following the same idea from Solidity here, to keep a separate 060 branch? We should merge master into master_060 once in a while, as it's done there.

test/smtcallback.js
leonardoalt: Right

leonardoalt:
Updated @erak 's comment.
@axic Anything else needed here?

axic:
It is failing tests? ("Invalid callback kind specified")

leonardoalt:
Yes, this is because it uses `soljson.js` from the latest release.
I think the merge order should be:
1) Merge https://github.com/ethereum/solidity/pull/7447 into Solidity `develop_060`. The external tests still do not pass there but I think I can write a tiny `solc-js` PR that will make it pass.
2) Change this PR to use `soljson.js` from `develop_060`. This would make the tests here pass as well.

leonardoalt:
Actually we can't make the tests here pass without making it download `soljson.js` from `develop_060`.

wrapper.js
```diff
@@ -26,25 +26,46 @@ function setupMethods (soljson) {
...
-  var runWithReadCallback = function (readCallback, compile, args) {
-    // Forward compatibility with 0.6.x
-    if (typeof readCallback === 'object') {
-      readCallback = readCallback.import;
+  var runWithCallbacks = function (callbacks, compile, args) {
+    var readCallback;
+    var smtSolverCallback;
+
+    if (callbacks !== undefined) {
+      readCallback = callbacks.importCallback;
```
axic: Actually we called this `import` in 0.5.x.

wrapper.js
```diff
@@ -71,21 +92,21 @@ function setupMethods (soljson) {
...
+      return runWithCallbacks(readCallback, compileInternal, [ input, optimize ]);
     };
   }
 
   var compileStandard = null;
   if ('_compileStandard' in soljson) {
     var compileStandardInternal = soljson.cwrap('compileStandard', 'string', ['string', 'number']);
     compileStandard = function (input, readCallback) {
-      return runWithReadCallback(readCallback, compileStandardInternal, [ input ]);
+      return runWithCallbacks(readCallback, compileStandardInternal, [ input ]);
```
axic: Hm, this will, fail, won't it?

smtsolver.js
```diff
@@ -3,14 +3,16 @@ var execSync = require('child_process').execSync;
...
 var potentialSolvers = [
   {
     name: 'z3',
-    params: '-smt2'
+    params: '-smt2 -t:' + timeout
   },
   {
     name: 'cvc4',
-    params: '--lang=smt2'
+    params: '--lang=smt2 --tlimit=' + timeout
```
axic: If there is a plan for another 0.5 release, then this timeout stuff could be merged into `master` first.

leonardoalt: Right

wrapper.js
leonardoalt: Updated

wrapper.js
axic: Actually realised the PR omitted the check for version. As we agreed two different ways are needed to pass callbacks based on the compiler version.

leonardoalt: True, this and the other 2 above it need the old `readWithReadCallback`

axic: All of them need it. For these above you can hardcode the old interface and make it version dependent under `solidity_compile`.

leonardoalt: Why not only these two? The others don't have the read callback.

axic: I meant all of the ones using `runWithCallbacks`.

test/compiler.js
```diff
@@ -4,6 +4,16 @@ const solc = require('../index.js');
 const linker = require('../linker.js');
 const execSync = require('child_process').execSync;
 
+function wrapImportCallbackVersion (solc, importCallback) {
+  var callback;
+  if (semver.gt(solc.semver(), '0.5.99')) {
+    callback = { import: importCallback };
+  } else {
+    callback = importCallback;
```
axic: Could just use `return` statements.

test/compiler.js
```diff
@@ -462,15 +476,15 @@ function runTests (solc, versionText) {
...
+        var output = JSON.parse(solc.compile(JSON.stringify(input), wrapImportCallbackVersion(solc, findImports)));
         var x = getBytecodeStandard(output, 'cont.sol', 'x');
         st.ok(typeof x === 'string');
         st.ok(x.length > 0);
         var L = getBytecodeStandard(output, 'lib.sol', 'L');
         st.ok(typeof L === 'string');
         st.ok(L.length > 0);
 
-        var outputNewApi = JSON.parse(solc.compile(JSON.stringify(input), { import: findImports }));
+        var outputNewApi = JSON.parse(solc.compile(JSON.stringify(input), wrapImportCallbackVersion(solc, findImports)));
```
axic: These two (`solc.compile`) should not use the wrapper, since the goal of their interface is to hide this.

test/compiler.js
```diff
@@ -224,7 +235,7 @@ function runTests (solc, versionText) {
           throw new Error('Could not implement this interface properly...');
         }
         st.throws(function () {
-          solc.lowlevel.compileCallback(JSON.stringify({sources: input}), 0, findImports);
+          solc.lowlevel.compileCallback(JSON.stringify({sources: input}), 0, wrapImportCallbackVersion(solc, findImports));
```
axic: 0.5 (and 0.6) don't have `solc.lowlevel.compileCallback`.

test/compiler.js
```diff
@@ -387,7 +398,10 @@ function runTests (solc, versionText) {
           }
         }
 
-        var output = JSON.parse(solc.lowlevel.compileStandard(JSON.stringify(input), findImports));
+        var output = JSON.parse(solc.lowlevel.compileStandard(
```
axic: This is the only case which should be needed.

test/smtcallback.js
```diff
@@ -0,0 +1,155 @@
...
+      sources: input
+    });
+    var tests = [
+      { cb: satCallback, expectations: ['Assertion violation happens here'] },
+      { cb: unsatCallback, expectations: [] },
+      { cb: errorCallback, expectations: ['BMC analysis was not possible'] }
+    ];
+    for (var i in tests) {
+      var test = tests[i];
+      var output = JSON.parse(solc.compileStandardWrapper(
```
axic: Use `solc.compile`. `compileStandard` and `compileStandardWrapper` are deprecated aliases (should have removed them in 0.5.2, but will do in 0.6)

test/smtcallback.js
```diff
@@ -0,0 +1,155 @@
...
+        { smtSolverCallback: test.cb }
+      ));
+      var errors = collectErrors(output);
+      st.ok(expectErrors(errors, test.expectations));
+    }
+    st.end();
+  });
+
+  t.test('Solidity smtCheckerTests', function (st) {
+    var testdir = './test/smtCheckerTests/';
```
axic: I think we should add a smoke test into this directory, otherwise none of this will be run on CI and it will be harder to find bugs introduced.

test/compiler.js
leonardoalt: Right

test/compiler.js
leonardoalt: Updated

test/compiler.js
leonardoalt: Ah, but the interface hides this based on the version of the used `soljson`. If you use the new `soljon`, this breaks without this version wrapper

test/smtcallback.js
leonardoalt: added

test/compiler.js
```diff
@@ -583,7 +587,6 @@ function runTests (solc, versionText) {
 
       t.test('compiling standard JSON (multiple sources on old compiler)', function (st) {
         var output = JSON.parse(solc.compile('{"language":"Solidity","sources":{"cont.sol":{"content":"import \\"lib.sol\\";"},"lib.sol":{"content":""}}}'));
-        console.log(output);
```
leonardoalt: @axic I'm not sure why this was there, was it intended?

axic: No, it was typo during merging.

PR #383, opened 2019-09-19
[0.6.0] Set up tests and fix read-only array length

erak:
Needed for https://github.com/ethereum/solidity/pull/7350.

This PR updates the workaround for breaking releases used by `npm test`. It also updates the newly added 0.6.0 tests to pass with `array.length` being an rvalue now.

### DIFF
--- a/test/DAO/DAO.sol
+++ b/test/DAO/DAO.sol
@@ -383,7 +383,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
             revert();
         lastTimeMinQuorumMet = now;
         minQuorumDivisor = 5; // sets the minimal quorum to 20%
-        proposals.length = 1; // avoids a proposal with ID 0 because it is used
+        proposals.push(); // avoids a proposal with ID 0 because it is used
 
         allowedRecipients[address(this)] = true;
         allowedRecipients[curator] = true;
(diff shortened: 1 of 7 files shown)


test/DAO/DAO.sol
```diff
@@ -447,8 +447,8 @@ contract DAO is DAOInterface, Token, TokenCreation {
         if (proposals.length == 1) // initial length is 1 (see constructor)
             lastTimeMinQuorumMet = now;
 
-        _proposalID = proposals.length++;
-        Proposal storage p = proposals[_proposalID];
+        Proposal storage p = proposals.push();
+        _proposalID = proposals.length;
```
leonardoalt: Since this is done after the `push` I think it should be `length - 1`

leonardoalt:
@erak shouldn't the target be `master_060`?

erak:
After the conversion on Friday, I thought we could try to get away without having a separate 0.6.0 branch. Would it be OK to have the new 0.6.0 determinism contracts in a 0.5.12 release of solc-js?

erak:
@leonardoalt https://github.com/ethereum/solidity/pull/7462 needs to be adjusted depending on the above as well.

leonardoalt:
> Would it be OK to have the new 0.6.0 determinism contracts in a 0.5.12 release of solc-js?

I think this would be a bit weird, wouldn't it? To me having a separate `master_060` feels more organized.
@axic ?

leonardoalt:
I'll merge for now, but let's write a PR next to extract the .sol tests to the Solidity repo.

PR #385, opened 2019-09-23
Backport new callback API to 0.5

axic:
Need to add a test case.

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -27,6 +27,11 @@ function setupMethods (soljson) {
 
   // This calls compile() with args || cb
   var runWithReadCallback = function (readCallback, compile, args) {
+    // Forward compatibility with 0.6.x
+    if (typeof readCallback === 'object') {
+      readCallback = readCallback.import;
+    }
+
     if (readCallback === undefined) {
       readCallback = function (path) {
         return {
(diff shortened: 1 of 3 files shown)


wrapper.js
```diff
@@ -33,6 +33,14 @@ function setupMethods (soljson) {
           error: 'File import callback not supported'
         };
       };
+    } else if (typeof readCallback === 'object') {
+      if (readCallback.imports === undefined) {
```
axic: Is "imports" the name we want to go with?

leonardoalt: I think that's fine.
Not super important, but to avoid a little code repetition you could do like https://github.com/ethereum/solc-js/pull/382/files#diff-2f0f3cebee013143afb9e2f50c940f8eR30 which assigns `readCallback` `undefined` before checking the argument's type, and then creates a fallback callback if it's still undefined.

axic: How about the simplification I've pushed now?

leonardoalt: Nice!

wrapper.js
```diff
@@ -27,6 +27,11 @@ function setupMethods (soljson) {
 
   // This calls compile() with args || cb
   var runWithReadCallback = function (readCallback, compile, args) {
+    // Forward compatibility with 0.6.x
+    if (typeof readCallback === 'object') {
+      readCallback = readCallback.imports;
```
axic: @leonardoalt @chriseth what should be the name of the callback?

leonardoalt: I'm fine with anything reasonable similar to that.
import, imports, read

axic:
@leonardoalt updated tests and readme.

Issue #361, opened 2019-05-11
Reduce project dependencies.


MicahZoltu:
This is a minor quality of life thing, but at the moment the NPM `solc` package brings with it 77 dependencies.  Given the simplicity of this project (just a wrapper around solc binaries) and the fact that almost all of the work with this library occurs at compile time, it feels like there is an opportunity to reduce the dependencies that `solc` brings into a project using it.

Can any of the dependencies be moved over to `devDependencies`?  Can any of the dependencies be replaced with just some inline code?

Looking at solc in http://npm.anvaka.com/#/view/2d/solc, it appears that the number one offender is `yargs` followed by `fs-extra` and `keccak`.  I already have a PR out for removing `keccak` (in favor of `js-sha3`), and it appears that `fs-extra` is used in one place and could easily be replaced with node's built-in `fs` package.  `yargs` I suspect will be harder to get rid of, though perhaps there is a similar library that doesn't bring in 58 transitive dependencies?

axic:
I'd happy to replace fs-extra, but yargs would be hard.

axic:
> the fact that almost all of the work with this library occurs at compile time

I'm not sure what you mean by that. The only piece which could be considered compile time is `downloadCurrentVersion.js`.

MicahZoltu:
Hmm, I'm not sure what I was thinking with that particular comment.  🤪

MicahZoltu:
Perhaps a solution to the yargs problem would be to have a separate package for the CLI from the library?  All of my projects compile programmatically (not using the CLI), and I don't think I'm alone in following this pattern.  Since `yargs` presumably is only used by the CLI portion, that would all go away if this project was split.

axic:
Do you want to submit a PR for replacing fs-extra? I assume we only use one or two helpers, and having it inline defined would not result in more than 20 extra lines of code?

MicahZoltu:
I'm considering it.  If someone beats me to it though that would make me happy.  😉

MicahZoltu:
If `fs-extra` was replaced by `fs` (built-in) then the only function missing would be `ensureDirSync`.  Looking at `fs-extra` it is just an alias of `makeDirSync` internally.  The following code I think is what would be required to remove the dependency on `fs-extra` entirely:
```js
function checkPath (pth) {
  if (process.platform === 'win32') {
    const pathHasInvalidWinCharacters = /[<>:"|?*]/.test(pth.replace(path.parse(pth).root, ''))

[...]

PR #391, opened 2019-10-27
Replaces yargs with commander.

MicahZoltu:
Works towards resolving #361.  Yargs brings in ~26 packages while commander brings in 1.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -11,48 +11,29 @@ var smtsolver = require('./smtsolver.js');
 // FIXME: remove annoying exception catcher of Emscripten
 //        see https://github.com/chriseth/browser-solidity/issues/167
 process.removeAllListeners('uncaughtException');
+var commander = require('commander');
 
-var yargs = require('yargs')
-  .usage('Usage: $0 [options] [input_file...]')
-  .option('version', {
-    describe: 'Show version and exit.',
-    type: 'boolean'
-  })
-  .option('optimize', {
-    describe: 'Enable bytecode optimizer.',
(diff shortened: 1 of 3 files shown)


MicahZoltu:
It is worth noting, I'm leaning on existing tests covering this.  I did do a run of `node ./solcjs --bin test/contracts/Smoke.sol` and it appeared to "do the right thing".  I also tried with `--abi` which also appears to have worked.

MicahZoltu:
Looking into test failures now (the tests won't run on my machine for some reason).

solcjs
```diff
@@ -11,48 +11,29 @@ var smtsolver = require('./smtsolver.js');
...
+
+var program = new commander.Command();
+program.name('solcjs');
+program.version(solc.version());
+program
+  .option('--version', 'Show version and exit.')
+  .option('--optimize', 'Enable bytecode optimizer.')
+  .option('--bin', 'Binary of the contracts in hex.')
+  .option('--abi', 'ABI of the contracts.')
+  .option('--standard-json', 'Standard JSON configuration file.')
```
axic: This doesn't take a file name so that the help text is confusing. I'd suggest to stick to the original text.

MicahZoltu: Oops, I have no idea why I changed that.  Fixed.

MicahZoltu: Ugh, I pushed the wrong branch... give me a few minutes while I clean up the mess I just created for myself.

MicahZoltu: Fixed.

PR #392, opened 2019-10-27
Updates scripts so that they run on Windows.

MicahZoltu:
For whatever reason, most of the packages used in scripts do not properly package cross-platform scripts (only exception is `tape`).  This change makes it so the scripts are run from node_modules rather than as commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,12 +7,12 @@
     "solcjs": "solcjs"
   },
   "scripts": {
-    "lint": "semistandard",
+    "lint": "node ./node_modules/semistandard/bin/cmd.js",
     "prepublish": "node downloadCurrentVersion.js && node verifyVersion.js",
     "pretest": "npm run lint",
     "test": "tape ./test/index.js",
-    "coverage": "istanbul cover node_modules/tape/bin/tape ./test/index.js",
-    "coveralls": "npm run coverage && coveralls <coverage/lcov.info"
+    "coverage": "node ./node_modules/istanbul/lib/cli.js cover ./node_modules/tape/bin/tape ./test/index.js",
+    "coveralls": "npm run coverage && node ./node_modules/coveralls/bin/coveralls.js <coverage/lcov.info"
   },
   "repository": {
     "type": "git",


axic:
I don't fully understand why this would be needed, the package `semistandard` clearly states it should be used without the need to mess with `node_module` paths.

MicahZoltu:
It is worth noting that I am not an expert in these matters and the below is how I understand the situation.  The original symptom of my specific problem is that I am unable to run any of the NPM scripts due to "command not found" errors on Windows, with the exception of being able to run a script that only references `tape` (see below on why `tape` works).

----

NPM CLI tools are (more or less) any package that you get from NPM that puts a file in the `node_modules/.bin` directory.  You can put whatever you want in there, and that folder is added to the path (effectively) when running an NPM script.  When an NPM script says `semistandard`, the OS will look for anexecutable on the path.  On Linux, it will find `node_modules/.bin/semistandard` and execute that.  On Windows however, it does not recognize that file as an executable file so it will not execute it and instead will throw an error.

The correct way to create cross-platform CLI packages on NPM is to drop both a `.cmd` file in `node_modules/.bin` and a Linux executable.  I believe there are tools built in to NPM that do this, though I don't have familiarity with using them.  If you take a look at the `node_modules/.bin` directory after doing `npm install` for this project you'll notice that only `tape` has both a `tape` file and a `tape.cmd` file in `node_modules/.bin`.  This means that the `tape` CLI _is_ cross-platform, while the others aren't.

In the case of the tools being used here, we are lucky in that the CLIs are really NodeJS applications, which means we can just call `node_modules/<package>/<script>` directly rather than using the named CLI script, so we can still use these tools cross-platform, even if their maintainers aren't publishing cross-platform runner scripts.

MicahZoltu:
I did a bit of digging, [it appears](https://stackoverflow.com/a/10398567/879046) that the problem is that those packages include a `.js` extension in their `package.json` `bin` property.  Apparently this causes NPM to not do-the-right-thing when installing the package on Windows.  If you look at `tape`'s `package.json`, you can see that they omit the `.js` in their script and it seems to cause NPM to do-the-right-thing.

axic:
Good job at figuring out the reasons. I think it would be nice to upstream the fixes, if you would have the time. (For coveralls and semistandard, istanbul is unmaintained and I'm trying to replace it with the successor in #373).

I'm not keen on making `package.json` even more worse to read, but I understand it could be a bad experience for the (small minority?) of Windows users.

PR #395, opened 2019-10-30
[WIP] Set version to 0.6.0

leonardoalt:


leonardoalt:
Still need to make it download `soljson.js` from Solidity `develop_060`.

PR #398, opened 2019-10-30
Backport new callback API to 0.5 to compileStandardWrapper

axic:
Fixes #385

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -109,6 +109,11 @@ function setupMethods (soljson) {
       });
     }
 
+    // Forward compatibility with 0.6.x
+    if (typeof readCallback === 'object') {
+      readCallback = readCallback.import;
+    }
+
     if (readCallback !== undefined && typeof readCallback !== 'function') {
       return formatFatalError('Invalid import callback supplied');
     }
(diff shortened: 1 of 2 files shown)


wrapper.js
```diff
@@ -109,6 +109,11 @@ function setupMethods (soljson) {
       });
     }
 
+    // Forward compatibility with 0.6.x
+    if (typeof readCallback === 'object') {
```
leonardoalt: At this point it shouldn't be an object, right? The only case that would support it is `compileStandard` which would have returned in the first lines

axic: The first line in `compilerStandardWrapper` is a short circuit to `compileStandard` if it exists. Handling of the callback is done there.

axic: We could move this single if there to avoid duplication of it, but not sure there is any point adding the smt callback into the wrapper anyway (as it will never work) so perhaps it makes more sense keeping the PR as is anticipating the smt changes will be merged.

leonardoalt: Ah sorry, confused it with `solidity_compile`

leonardoalt: But still, this condition would never be true, since the only case that actually accepts an object with callbacks (`solidity_compile`) would lead to `compileStandard` existing

leonardoalt: Ah, or would that be for new solc-js + old solc?

axic: > Ah, or would that be for new solc-js + old solc?

Yes

PR #404, opened 2019-11-09
Reorder functions in wrapper (move version/license early)

axic:
In preparation for #382


axic:
@leonardoalt this actually isn't needed due to scopes in JS (e.g. no reordering is nedeed), but makes it easier to read I guess.

PR #415, opened 2019-11-26
Change solcjs to be async

axic:


r0qs:
Hi, I will be closing this PR for now. We would like to make the solc-js cli async eventually, but this is not a priority, and this PR has been stale for too long. I will make an issue for this, so it does get forgotten.

PR #416, opened 2019-11-26
Export load and loadRemoteVersion without loading a compiler

axic:
The goal of this change is to load a remote compiler version without loading a local one.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,3 @@
 const wrapper = require('./wrapper.js');
 
-module.exports = wrapper(require('./soljson.js'));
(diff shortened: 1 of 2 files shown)


leonardoalt:
Looks good, tests failing though

axic:
This is an ancient commit, but will fix this once 0.6.0 is done.

r0qs:
Hi, it would be great to split the local and remote load logic. However, this is a breaking change that directly affects other projects that already use this feature, like Remix. We will not be doing it for now. But I'll create an issue to remind us to finish this eventually ;)

PR #418, opened 2019-11-28
Update wrapper to support newer Emscripten (1.39.3)

axic:


ekpyron:
For reference: needed, if we want to go ahead with https://github.com/ethereum/solidity/pull/7832 at some point, but shouldn't have backwards compatibility issues until then.

wrapper.js
```diff
@@ -28,22 +28,27 @@ function setupMethods (soljson) {
...
 
-  var copyString = function (str, ptr) {
+  var copyToCString = function (str, ptr) {
     var length = soljson.lengthBytesUTF8(str);
     var buffer = soljson._malloc(length + 1);
     soljson.stringToUTF8(str, buffer, length + 1);
     soljson.setValue(ptr, buffer, '*');
   };
 
+  var copyFromCString = function (ptr) {
```
ekpyron:
```suggestion
  var copyFromCString = (soljson.UTF8ToString || soljson.Pointer_stringify);
```

PR #419, opened 2019-11-28
Increase code coverage for older compiler versions

axic:
Part of #180.

Todo:
- [x] test for single file
- [x] test abstract contracts
- [x] test for compiler warnings
- [ ] test for compiler errors
- [x] test translation of gas estimation
- [ ] test with every single release...

PR #421, opened 2019-11-29
Add tests for gas estimation

axic:
Pulled from #419.


test/compiler.js
```diff
@@ -35,6 +35,20 @@ function runTests (solc, versionText) {
     }
   }
 
+  function getGasEstimate (output, fileName, contractName) {
```
axic: In-line with the above, which says `getBytecodeStandard`.

PR #427, opened 2019-12-09
Implement support for the new memory allocation API

axic:


wrapper.js
```diff
@@ -134,6 +153,14 @@ function setupMethods (soljson) {
       throw e;
     }
     removeFunction(cb);
+    if (isVersion6) {
```
axic: Actually could run `reset` in every case, as it should not have an adverse effect on pre-0.6.0.

wrapper.js
```diff
@@ -31,12 +31,30 @@ function setupMethods (soljson) {
...
-    // Assuming copyToCString is only used in the context of wrapCallback, solc will free these pointers.
-    // See https://github.com/ethereum/solidity/blob/v0.5.13/libsolc/libsolc.h#L37-L40
-    var buffer = soljson._malloc(length + 1);
+    //
+    // Before 0.6.0:
+    //   Assuming copyToCString is only used in the context of wrapCallback, solc will free these pointers.
+    //   See https://github.com/ethereum/solidity/blob/v0.5.13/libsolc/libsolc.h#L37-L40
+    //
+    // After 0.6.0:
+    //   The duty is on solc-js to free these pointers. We accomplish that by calloed `reset` at the end.
```
axic: Typo: calloed -> calling

wrapper.js
```diff
@@ -31,12 +31,30 @@ function setupMethods (soljson) {
     };
   }
 
+  var alloc;
+  if ('_solidity_alloc' in soljson) {
+    alloc = soljson.cwrap('solidity_alloc', 'number', [ 'number' ]);
+  } else {
+    // pre 0.6.0
+    alloc = soljson._malloc;
```
leonardoalt: Maybe assert that it exists (if applicable)?

wrapper.js
```diff
@@ -61,6 +79,8 @@ function setupMethods (soljson) {
   var wrapCallbackWithKind = function (callback) {
     assert(typeof callback === 'function', 'Invalid callback specified.');
     return function (context, kind, data, contents, error) {
+      // Must be a null pointer.
```
leonardoalt: Why? Just now?

wrapper.js
leonardoalt: Yea, just need to create it also for `!isVersion6`

wrapper.js
```diff
@@ -31,13 +31,41 @@ function setupMethods (soljson) {
...
+    //   Assuming copyToCString is only used in the context of wrapCallback, solc will free these pointers.
+    //   See https://github.com/ethereum/solidity/blob/v0.5.13/libsolc/libsolc.h#L37-L40
+    //
+    // After 0.6.0:
+    //   The duty is on solc-js to free these pointers. We accomplish that by calling `reset` at the end.
+    var buffer;
+    if (isVersion6) {
+      buffer = alloc(length + 1);
+      soljson.stringToUTF8(str, buffer, length + 1);
+      resize(buffer, length);
```
chriseth: Can you explain why you need the resize? Is it really worth saving a single byte?

axic: See https://github.com/ethereum/solidity/pull/7898#issuecomment-563482425

chriseth: It is a c interface, so we should consider this to have a trailing zero. I think it is the responsibility of the C++ part to not take the length property for granted but instead search for the zero terminator.

axic: That is listed as an option on the issue. Anyway this change confirms that the interface is working well otherwise (since all tests pass).

We can agree on the final API tomorrow on the main issue.

wrapper.js
```diff
@@ -31,13 +31,35 @@ function setupMethods (soljson) {
...
+    //
+    // Before 0.6.0:
+    //   Assuming copyToCString is only used in the context of wrapCallback, solc will free these pointers.
+    //   See https://github.com/ethereum/solidity/blob/v0.5.13/libsolc/libsolc.h#L37-L40
+    //
+    // After 0.6.0:
+    //   The duty is on solc-js to free these pointers. We accomplish that by calling `reset` at the end.
+    var buffer;
+    if (isVersion6) {
+      buffer = alloc(length + 1);
```
ekpyron: To complement my comment on the solidity PR: We can just use ``alloc(length)`` here, while still passing ``length + 1`` to ``stringToUTF8`` below and it's guaranteed to be valid, while then the ``std::string`` on the solidity side won't contain an additional zero.

wrapper.js
axic: Yeah forgot that it wasn't exposed :wink:

wrapper.js
axic: It was an oversight. It helps catching cases when the API changed but we didn't update it here. Otherwise it doesn't do anything.

wrapper.js
axic: Most of the relevant discussion took place on https://github.com/ethereum/solidity/pull/7898

wrapper.js
axic: Added.

PR #441, opened 2020-01-13
Workarounds for running tests against wasm rebuilds of old compilers.

ekpyron:
@axic I used this branch for running the solc-js tests against the wasm rebuilds now... additionally I patched ``compiler.js`` on the fly to replace ``'latest'`` with the respective version string...

Making this a draft, since we'll probably need to discuss if and how this should end up in master.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -92,7 +92,10 @@ tape('CLI', function (t) {
     spt.stdin.on('finish', function () {
       spt.stderr.empty();
       spt.stdout.match(/Contract.sol/);
-      spt.stdout.match(/userdoc/);
+      if (semver.gt(pkg.version, '0.4.6')) {
+        // TODO: is it legit to skip this for <=0.4.6?
+        spt.stdout.match(/userdoc/);
+      }
       spt.succeeds();
       spt.end();
     });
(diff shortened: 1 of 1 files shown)


axic:
@ekpyron ah we're not testing the CLI with different versions though.

ekpyron:
Not in the "normal" solc-js test suite... what I do in https://github.com/ethereum/solidity/blob/98ec970813c47ebbb3ad9b11042116de5b89aac5/scripts/wasm-rebuild/docker-scripts/rebuild_tags.sh#L158 (part of https://github.com/ethereum/solidity/pull/8118) is to copy the "new old" ``soljson.js`` over to test it - that's how I noticed that they don't run without these changes.

But yeah: we need to decide what to do about this - if we make the CLI tests compatible with old versions, then we should probably always run them against old versions as well...

Related to this: I'd also vote for making it possible to run the test suite against one particular soljson.js version, resp. even better against a particular file only without needing to edit the test suite for it.

ekpyron:
Not sure how difficult it would be to let the npm/tape stuff take a soljson.js file as optional argument or something like that... my experience with them is only tangential...

PR #442, opened 2020-01-21
Migrating to Typescript

edisinovcic:
This PR is the migration of the whole library from javascript to typescript.

Jest testing framework is being used in rewrite instead of tape.

TODO: 
* check if the wrapper is okay in this format
* fix coverage:jest problem with memory leakage (for now not part of the CI)


__tests__/compiler.spec.ts
```diff
@@ -0,0 +1,853 @@
...
+          }
+  
+          let output = JSON.parse(solc.lowlevel.compileSingle('contract x { this is an invalid contract }'));
+          if (semver.lt(solc.semver(), '0.1.4')) {
+            expect(output.error.indexOf('Parser error: Expected identifier')).not.toBe(-1);
+            return;
+          }
+
+          // TODO: what plan does? https://github.com/substack/tape#tplann
+          //st.plan(3);
```
edisinovcic: @chriseth does this mean that tests run 3 times?
https://github.com/substack/tape#tplann 

Definition on the github:
"Declare that n assertions should be run. t.end() will be called automatically after the nth assertion. If there are any more assertions after the nth, or after t.end() is called, they will generate errors."

chriseth: Yes

chriseth: Actually no, it says that there are 3 checks to be run. Because they are async, you basically tell the test framework when to stop waiting for more checks.

edisinovcic:
@chriseth changes to .js files are due to semistandard, on the compiler.ts part all tests are running and it should be the same as compiler.js with types added. 

It's a big change so please let me know if I missed something.

edisinovcic:
@axic please take a look at this when you have time. Tnx ;)

src/wrapper.ts
```diff
@@ -0,0 +1,368 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import assert = require('assert');
+import translate = require('./translate');
+import requireFromString = require('require-from-string');
+import https = require('https');
+import MemoryStream = require("memorystream");
+import semver = require('semver');
+import { IncomingMessage } from 'http';
+
+export class Wrapper {
```
edisinovcic: @axic Can I leave as a class or rather revert it to the setupMethods as it was before?

MambaDev:
@edisinovcic You probably should update the gitignore to include the .nyc_output/ path, since these should not be included within git.

edisinovcic:
> @edisinovcic You probably should update the gitignore to include the .nyc_output/ path, since these should not be included within git.

Tnx will do. I hope that I will have some time to get back too this in near future to finish it completely. Tnx for letting me know!

franco-roura:
Would love to see this one merged, let me know if there's anything I can do to help 😊

cameel:
@franco-roura I took a quick look at the PR and for me the biggest issue is that it looks pretty much like a complete rewrite of the library. Tests are being rewritten too so it's pretty much impossible to be sure if all the functionality is preserved. Reviewing this is going to be nightmarish and I suspect it would take more time than it took to write this :)

If you want to push this forward, then I think that the best course of action would be to find a way to do this piecemeal, in a set of smaller PRs. For example the switch to a different test framework could probably be done on its own, without modifying the code it tests. The PR adds some new dependences - for example ESLint looks like something that can be added in a separate PR. Finally, since TypeScript is supposed to be backwards-compatible with JS, maybe we could first convert the project to TS with as little actual JS changes as possible and only after that add type annotations? This PR only adds new .ts files, without removing the .js equivalents so it's not even clear to me how much it modifies the original code.

Apart from the PR being big and unwieldy, you can probably see from the list of old, outdated PRs in this repo that we do not have enough manpower for reviewing these.  The Solidity team consists mostly of C++ developers and most of the effort goes into the compiler. Any help in getting these PRs reviewed to the point where the biggest problems are ironed out would likely speed things up more than writing more code.

r0qs:
Hi, thank you very much for the PR. But as pointed out, reviewing a PR that makes so many changes at once is not practical. However, this motivated others to contribute with small changes inspired by this PR, so thanks for your effort!

Part of your suggestions are already merged, and some are still in progress. If there are still things that you would like to add to solc-js, please feel free to open a new PR :D

PR #446, opened 2020-02-18
Use GitHub raw.githubusercontent.com to avoid rate limiting

### LINKED ISSUE #445: Install often times out due to github pages rate limit
We're having trouble building contracts on our CI machine. The build fails very often with a rate limiting error:

```
 | FetchError: response timeout at https://ethereum.github.io/solc-bin/bin/soljson-v0.4.24+commit.e67f0147.js over limit: 20000
 |     at Timeout._onTimeout (/chainlink/node_modules/node-fetch/lib/body.js:117:12)
```

Is there some way we can provide a cached version of the soljson binaries, or supply a different url to work around this?

jleeh:
Fixes #445.

- Rate limit is being hit causing flakey builds for anyone who depends on solc-js remote downloading in their software. The change to GitHub raw URLs avoids hitting rate limits.
- ~~Ability to override the repository all together with an environment variable `SOLC_REPO`. Allowing people to use alternative repos, including internal.~~

### DIFF
--- a/downloadCurrentVersion.js
+++ b/downloadCurrentVersion.js
@@ -13,7 +13,7 @@ function getVersionList (cb) {
   console.log('Retrieving available version list...');
 
   var mem = new MemoryStream(null, { readable: false });
-  https.get('https://ethereum.github.io/solc-bin/bin/list.json', function (response) {
+  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json', function (response) {
     if (response.statusCode !== 200) {
       console.log('Error downloading file: ' + response.statusCode);
       process.exit(1);
@@ -40,7 +40,7 @@ function downloadBinary (outputName, version, expectedHash) {
   });
 
   var file = fs.createWriteStream(outputName, { encoding: 'binary' });
-  https.get('https://ethereum.github.io/solc-bin/bin/' + version, function (response) {
+  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/' + version, function (response) {
     if (response.statusCode !== 200) {
       console.log('Error downloading file: ' + response.statusCode);
       process.exit(1);
--- a/wrapper.js
+++ b/wrapper.js
@@ -328,7 +328,7 @@ function setupMethods (soljson) {
     // instead of from the local filesystem.
     loadRemoteVersion: function (versionString, cb) {
       var mem = new MemoryStream(null, {readable: false});
-      var url = 'https://ethereum.github.io/solc-bin/bin/soljson-' + versionString + '.js';
+      var url = 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/soljson-' + versionString + '.js';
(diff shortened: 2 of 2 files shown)


downloadCurrentVersion.js
```diff
@@ -13,7 +15,7 @@ function getVersionList (cb) {
   console.log('Retrieving available version list...');
 
   var mem = new MemoryStream(null, { readable: false });
-  https.get('https://ethereum.github.io/solc-bin/bin/list.json', function (response) {
+  https.get(SOLC_REPO + 'list.json', function (response) {
```
chriseth:
```suggestion
  https.get(SOLC_REPO + '/list.json', function (response) {
```
Just in case people forget a trailing slash in the env variable?

chriseth:
Can you document this feature in the readme?

@edisinovcic does this conflict with your work?

edisinovcic:
No, it's definitely a welcomed PR.

downloadCurrentVersion.js
```diff
@@ -3,6 +3,8 @@
 // This is used to download the correct binary version
 // as part of the prepublish step.
 
+const SOLC_REPO = process.env.SOLC_REPO || 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/';
```
axic: What is the use case for changing the official URL here?

axic:
Changing it here, while keeping the old URL in `wrapper.js` creates an inconsistency in the library.

edisinovcic:
Definitely @axic but I can update it for the whole project for it to be configurable if that's okay with you?

axic:
> for the whole project for it to be configurable

I am not yet seeing the use case for that. Also note that `wrapper.js` should definitely not use environment variables as it is a library and not CLI.

downloadCurrentVersion.js
jleeh: The GitHub pages URL becomes rate limited once you exceed certain bandwidth allowance, which made our CI builds fail when downloading solc. Using the raw URL bypasses the pages rate limits.

axic: I understand that, but not the reason for `env.SOLC_REPO`.

jleeh:
>I am not yet seeing the use case for that.

Mostly for pointing to internal repositories/elsewhere. Even though there's checksums, some teams could want to lock versions in their own repository, or remove the dependency on GitHub all together.

I'll update the URL in wrapper.js, but in terms of changing it since it's a lib, I feel I wouldn't be best making that change.

downloadCurrentVersion.js
jleeh: Commented underneath.

axic:
I'd prefer we merge this without the override (to include it in 0.6.3) and then we can discuss the override separately and land it in 0.6.4. Does that sounds like a good middle gorund?

axic:
@jleeh can you create a new PR for the override?

jleeh:
Sure. I'll base it on this branch to avoid future conflicts.

PR #447, opened 2020-02-18
 Ability to override remote download URL with env var

jleeh:
When using the CLI/high-level API, add the ability to override the remote download URL with the `SOLC_REPO` environment variable.

### DIFF
--- a/downloadCurrentVersion.js
+++ b/downloadCurrentVersion.js
@@ -13,7 +13,7 @@ function getVersionList (cb) {
   console.log('Retrieving available version list...');
 
   var mem = new MemoryStream(null, { readable: false });
-  https.get('https://ethereum.github.io/solc-bin/bin/list.json', function (response) {
+  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json', function (response) {
     if (response.statusCode !== 200) {
       console.log('Error downloading file: ' + response.statusCode);
       process.exit(1);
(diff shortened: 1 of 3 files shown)


downloadCurrentVersion.js
```diff
@@ -3,6 +3,8 @@
 // This is used to download the correct binary version
 // as part of the prepublish step.
 
+const SOLC_REPO = process.env.SOLC_REPO || 'https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin';
```
axic: My main concern with this override is that compiler hashes are not enforced. This means it would be possible that maliciously someone keeps downloading tainted compiler versions when using truffle, for example.

Is the use of private repositories of solc-bin a common thing? Is it still needed after the change to raw.githubusercontent?

A better solution would be to enforce hashes of downloaded compilers, but then we'd need to bake in the hash of `list.json` tied to a given version and nightlies could not be downloaded (well it would be flagged as untrusted).

jleeh: >My main concern with this override is that compiler hashes are not enforced.
My thinking was to enforce compiler hashes directly within your control to stop them being tainted. For example, if I was to setup an internal solc internal repo, all I would do is directly copy `solc-bin` and host it elsewhere. There's still potential when using GitHub that the maintainers of the repo update previous checksums and compiler versions.

>Is the use of private repositories of solc-bin a common thing? Is it still needed after the change to raw.githubusercontent?
I'd say it's common in the corporate world, but I've not seen it too much in ETH dev? For example, how tools like Artifactory can mirror RPM/npm repos to be used internally or simply just as a cache to be more redundant.

Overall, I think the main benefit of being able to use overrides is that teams could use custom solc versions which they host internally then hook in to tools like Truffle.

axic: > Overall, I think the main benefit of being able to use overrides is that teams could use custom solc versions which they host internally then hook in to tools like Truffle.

I'm not sure we should encourage the use of unofficial compiler versions. Do you have a reason in mind for an unofficial compiler build? The project has always tried to be accommodating to external input, which has led to avoiding forks of the compiler.

r0qs:
Hi, thanks for the PR. We think that allowing the download from different sources may be relevant. However, it should be from a trusted list of sources for security reasons.

I will be closing this PR for now since it does not accomplish something that we want now. If you have any new ideas on performing such an operation more trustfully, please feel free to open a new PR with your suggestions.

PR #455, opened 2020-03-12
Add test option to disable remote version tests.

ekpyron:
We can do this differently, but we should really have something like this. It's extremely annoying for running the tests from the solidity repo, that you can't force it to only run against the version of ``soljson.js`` you want...

### DIFF
--- a/test/compiler.js
+++ b/test/compiler.js
@@ -4,6 +4,8 @@ const solc = require('../index.js');
 const linker = require('../linker.js');
 const execSync = require('child_process').execSync;
 
+var noRemoteVersions = (process.argv.indexOf('--no-remote-versions') >= 0);
+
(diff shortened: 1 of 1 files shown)


leonardoalt:
Well, how would you apply this in the Solidity repo? Simply not use the remote versions here?

ekpyron:
In solidity in ``test/externalTests/solc-js/solc-js.sh`` we'd just need to change
``npm test`` to ``npm run test -- --no-remote-versions``.

ekpyron:
By the way: I'd still find this useful and see no harm in it. So far I still have to comment these parts out manually all the time...

ekpyron:
By the way almost a year later I'd *still* like to have this :-D.

chriseth (approved):

PR #467, opened 2020-05-15
Added basePath

uspasojevic96:


christianparpart:
Should be merged *after* https://github.com/ethereum/solidity/pull/8927 was merged.

solcjs
```diff
@@ -22,6 +22,7 @@ program
   .option('--bin', 'Binary of the contracts in hex.')
   .option('--abi', 'ABI of the contracts.')
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
+  .option('--base-path', 'Use the given path as the root of the source tree instead of the root of the filesystem.')
```
chriseth, line 25: Description is ambiguous (what is the root of what?).
Maybe `Specifies the root of the source tree (default value: "/").`?

uspasojevic96: You are right, changed it.

solcjs
```diff
@@ -93,7 +94,13 @@ var output = JSON.parse(solc.compile(JSON.stringify({
     }
   },
   sources: sources
-})));
+}
+
+if (program.basePath) {
+  options.settings.basePath = basePath
```
chriseth, line 100: I don't think this is as easy as just forwarding the option. We have to modify the read callback.

PR #470, opened 2020-06-04
Pass the wasm signature to addFunction.

ekpyron:
This should be backwards compatible just as it is, shouldn't it? Let's see what the tests say.
Relates to https://github.com/ethereum/solidity/pull/9111

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -140,7 +140,7 @@ function setupMethods (soljson) {
     var addFunction = soljson.addFunction || soljson.Runtime.addFunction;
     var removeFunction = soljson.removeFunction || soljson.Runtime.removeFunction;
 
-    var cb = addFunction(singleCallback);
+    var cb = addFunction(singleCallback, 'viiiii');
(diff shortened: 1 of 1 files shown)


christianparpart:
@ekpyron I know this PR is merged already and nobody seemed to have questioned it. But why do you pass 6 times `i` here? `v` is for return void type, and then it's followed by a sequence encoding the function parameter types, where `i` stands for int32. But why six times?

cameel:
@christianparpart
https://github.com/ethereum/solc-js/blob/16ac5ab6a8f8ccca2a36028e4341a591e708d0b0/wrapper.js#L123-L143
https://github.com/ethereum/solc-js/blob/16ac5ab6a8f8ccca2a36028e4341a591e708d0b0/wrapper.js#L79-L92

The original `singleCallback` takes 2 args but it's passed through `wrapCallbackWithKind()`, which returns a wrapper that actually expects 5 args (it's 5, not 6).

Issue #169, opened 2017-12-22
Warn as error in solcjs


posix4e:
Is it reasonable to enable an option where warnings return a false error code?

axic:
Since this `solcjs` CLI should match with the upstream `solc` binary, I suggest we wait for a decision on https://github.com/ethereum/solidity/issues/10277.

PR #476, opened 2020-07-05
Add option to return warning as error

t4sk:
Option to return warning as error

https://github.com/ethereum/solc-js/issues/169


r0qs:
Hi, @t4sk thanks for the PR, but the `solc-js` cli should match the behavior of the `solc`, which currently, does not return warnings as errors. So I will be closing this PR for now. Thanks for your contribution, though.

PR #478, opened 2020-07-15
From 0.7.0 on CHC reports assertion failures ingoring the callback

leonardoalt:
In https://github.com/ethereum/solidity/pull/9352, CHC reports assertion failures without considering the callback results, since there is no CHC portfolio the same way there is an SMTPortfolio. So we either need this PR or a mechanism that allows us to choose the engine, where we could choose BMC only and not have the extra if here.

### DIFF
--- a/test/smtcallback.js
+++ b/test/smtcallback.js
@@ -66,18 +66,26 @@ tape('SMTCheckerCallback', function (t) {
       sources: input
     });
 
-    // Solidity 0.6.9 comes with z3
     var tests;
     if (!semver.gt(solc.semver(), '0.6.8')) {
+      // Up to version 0.6.8 there were no embedded solvers.
       tests = [
         { cb: satCallback, expectations: ['Assertion violation happens here'] },
         { cb: unsatCallback, expectations: [] },
         { cb: errorCallback, expectations: ['BMC analysis was not possible'] }
       ];
-    } else {
+    } else if (!semver.gt(solc.semver(), '0.6.12')) {
+      // Solidity 0.6.9 comes with z3.
       tests = [
         { cb: satCallback, expectations: ['Assertion violation happens here'] },
         { cb: unsatCallback, expectations: ['At least two SMT solvers provided conflicting answers. Results might not be sound.'] },
+        { cb: errorCallback, expectations: ['Assertion violation happens here'] }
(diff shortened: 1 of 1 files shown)


leonardoalt:
Converted to draft because we need to wait for Solidity's release today

test/smtcallback.js
```diff
@@ -66,20 +66,27 @@ tape('SMTCheckerCallback', function (t) {
...
-    // Solidity 0.6.9 comes with z3
     var tests;
     if (!semver.gt(solc.semver(), '0.6.8')) {
       tests = [
         { cb: satCallback, expectations: ['Assertion violation happens here'] },
         { cb: unsatCallback, expectations: [] },
         { cb: errorCallback, expectations: ['BMC analysis was not possible'] }
       ];
-    } else {
+    // Solidity 0.6.9 comes with z3
```
chriseth, line 76: Shouldn't the comment befurther up?

leonardoalt: I moved it down here because this is the block that refers to 0.6.9. The block above has the old old behavior, <0.6.9.

test/smtcallback.js
```diff
@@ -66,20 +66,27 @@ tape('SMTCheckerCallback', function (t) {
...
+    } else if (!semver.gt(solc.semver(), '0.7.0')) {
```
leonardoalt, line 77: I think this is actually wrong. This should refer to the last 0.6 version

leonardoalt:
Updated

leonardoalt:
@chriseth @axic can we merge?

test/smtcallback.js
```diff
@@ -66,20 +66,28 @@ tape('SMTCheckerCallback', function (t) {
       sources: input
     });
 
-    // Solidity 0.6.9 comes with z3
     var tests;
     if (!semver.gt(solc.semver(), '0.6.8')) {
+      // Up to version 0.6.8 there were no embedded solvers.
```
chriseth, line 71: Oh it's "not greater than" - easy as cake!

axic: Should we change these to `lte`?

leonardoalt: Prolly

PR #479, opened 2020-07-20
Getting soljson.js via dedicated solc-bin.ethereum.org domain

cameel:
Currently `solc-js` is not using [solc-bin.ethereum.org](https://solc-bin.ethereum.org) domain to get `soljson.js` binaries from `solc-bin` repository. Instead it fetches the files from [raw.githubusercontent.com](https://raw.githubusercontent.com) URLs. This prevents us from freely reorganizing the structure of `solc-bin` since, unlike Github Pages, it does not provide the original file content when you access it via a symlink. Instead you get a text file with a path and `sojc-js` can't handle that currently (https://github.com/ethereum/solc-bin/pull/43, https://github.com/ethereum/solc-bin/pull/44).

Not using a domain under our control is also an obstacle to transparently moving files to a different hosting and this is exactly what we're facing now (see https://github.com/ethereum/solidity/issues/9258). We can fix it in newer versions of `solc-js` but the old ones will keep trying to get them from the old location.

This PR fixes the immediate issue by switching the URLs to [solc-bin.ethereum.org](https://solc-bin.ethereum.org). I originally assumed this to be just an oversight but after browsing through issues and PRs in this repo I see that the switch to [raw.githubusercontent.com](https://raw.githubusercontent.com) was actually intentional (#446, #445). It seems that the reason was that GH pages are rate-limited. This means that switching to the official domain would bring the problem back as it's currently pointing at GH pages too. I'm in the process of preparing an Amazon S3 bucket to mirror `solc-bin` content and I think I'll have to keep this PR as a draft until that's ready and the official domain is pointing at it.

### DIFF
--- a/downloadCurrentVersion.js
+++ b/downloadCurrentVersion.js
@@ -13,7 +13,7 @@ function getVersionList (cb) {
   console.log('Retrieving available version list...');
 
   var mem = new MemoryStream(null, { readable: false });
-  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json', function (response) {
+  https.get('https://solc-bin.ethereum.org/bin/list.json', function (response) {
     if (response.statusCode !== 200) {
       console.log('Error downloading file: ' + response.statusCode);
       process.exit(1);
@@ -40,7 +40,7 @@ function downloadBinary (outputName, version, expectedHash) {
   });
 
   var file = fs.createWriteStream(outputName, { encoding: 'binary' });
-  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/' + version, function (response) {
+  https.get('https://solc-bin.ethereum.org/bin/' + version, function (response) {
     if (response.statusCode !== 200) {
       console.log('Error downloading file: ' + response.statusCode);
       process.exit(1);
--- a/test/compiler.js
+++ b/test/compiler.js
@@ -813,7 +813,7 @@ const versions = [
 ];
 for (var version in versions) {
   version = versions[version];
-  execSync(`curl -o /tmp/${version}.js https://solc-bin.ethereum.org/bin/soljson-${version}.js`);
+  execSync(`curl -L -o /tmp/${version}.js https://solc-bin.ethereum.org/bin/soljson-${version}.js`);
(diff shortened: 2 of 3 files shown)


downloadCurrentVersion.js
```diff
@@ -13,7 +13,7 @@ function getVersionList (cb) {
   console.log('Retrieving available version list...');
 
   var mem = new MemoryStream(null, { readable: false });
-  https.get('https://raw.githubusercontent.com/ethereum/solc-bin/gh-pages/bin/list.json', function (response) {
+  https.get('https://solc-bin.ethereum.org/bin/list.json', function (response) {
```
cameel, line 16: `https.get()` won't handle HTTP redirects as far as I know. Unfortunately it looks like you need an external library for that. Any suggestions? Is it ok to add a new dependency just for that?

chriseth: Oh really? Yeah, please either add the most lightweight library or write a simple function that follows at most 3 redirect or something.

chriseth:
Oh this is really annoying. Can we maybe find a solution where we first try one url and then fall back to the other if it fails? We could even handle the "non-working symlinks" in such a way.

cameel:
@chriseth I just tried to reproduce the rate limiting issue to be able to test the fallback and manged to make a ton of requests without actually running into it.

I did 1k requests to `solc-bin.ethereum.org`, 1k to `ethereum.github.io` and later again 1.6k to `solc-bin.ethereum.org`. It was in total about 1.5 GB worth of `list.json` which is a non-negligible fraction of the monthly bandwidth cap for GH pages (50 GB). When I was working with github API earlier I was running into HTTP 403 errors after something like 50-100 requests so even if there's any rate limiting, it seems pretty reasonable.

So I'm not sure if the fallback is really needed given that we're switching to S3 anyway. I think that using the `ethereum.github.io` URL will not affect individual users and heavy usage in CI will just go back to being flaky for a short time until S3 is ready.

downloadCurrentVersion.js
cameel: I'll do it in a separate PR not to block this one. Right now we don't have any redirects in `solc-bin` anyway.

cameel:
I just noticed that it was still a draft. Switching to normal PR.

PR #484, opened 2020-08-10
Fix test expectation

leonardoalt:
This is needed for the cases where z3 is nondeterministic (unfortunately)


test/smtcallback.js
```diff
@@ -34,6 +34,8 @@ function expectErrors (errors, expectations) {
     if (errors[i].includes('Error trying to invoke SMT solver') || expectations[i].includes('Error trying to invoke SMT solver')) {
       continue;
     }
+    errors[i] = errors[i].replace('happens here.', 'happens here');
+    expectations[i] = expectations[i].replace('happens here.', 'happens here');
```
ekpyron, line 38: This one would be enough actually.

PR #506, opened 2020-12-07
Prepare SMT comparison for counterexamples

leonardoalt:
Needed by https://github.com/ethereum/solidity/pull/10477

### DIFF
--- a/test/smtcallback.js
+++ b/test/smtcallback.js
@@ -25,7 +25,7 @@ function collectErrors (solOutput) {
   return errors;
 }
 
-function expectErrors (errors, expectations) {
+function expectErrors (expectations, errors, ignoreCex) {
   if (errors.length !== expectations.length) {
     return false;
   }
@@ -34,8 +34,19 @@ function expectErrors (errors, expectations) {
     if (errors[i].includes('Error trying to invoke SMT solver') || expectations[i].includes('Error trying to invoke SMT solver')) {
       continue;
     }
-    expectations[i] = expectations[i].replace('happens here.', 'happens here');
-    if (!errors[i].includes(expectations[i])) {
+    // Expectations containing counterexamples might have many '\n' in a single line.
+    // These are stored escaped in the test format (as '\\n'), whereas the actual error from the compiler has '\n'.
+    // Therefore we need to replace '\\n' by '\n' in the expectations.
+    // Function `replace` only replaces the first occurrence, and `replaceAll` is not standard yet.
+    // Replace all '\\n' by '\n' via split & join.
+    expectations[i] = expectations[i].split('\\n').join('\n');
+    if (ignoreCex) {
+      expectations[i] = expectations[i].split('\nCounterexample')[0];
+      errors[i] = errors[i].split('\nCounterexample')[0];
+    }
+    // `expectations` have "// Warning ... " before the actual message,
+    // whereas `errors` have only the message.
+    if (!expectations[i].includes(errors[i])) {
       return false;
     }
   }
@@ -157,7 +168,8 @@ tape('SMTCheckerCallback', function (t) {
       }
       tests[sources[i]] = {
         expectations: expected,
-        solidity: { test: { content: preamble + source } }
+        solidity: { test: { content: preamble + source } },
+        ignoreCex: source.includes('// SMTIgnoreCex: yes')
       };
     }
 
@@ -201,7 +213,7 @@ tape('SMTCheckerCallback', function (t) {
       }
 
       // Compare expected vs obtained errors
-      st.ok(expectErrors(test.expectations, test.errors));
+      st.ok(expectErrors(test.expectations, test.errors, test.ignoreCex));
     }
 
     st.end();


test/smtcallback.js
```diff
@@ -146,6 +147,11 @@ tape('SMTCheckerCallback', function (t) {
         continue;
       }
 
+      if (source.includes('// SMTIgnoreCex: yes')) {
+        st.skip('Test ignores counterexamples.');
+        continue;
+      }
```
ekpyron, line 153: So you don't bother to just ignore the counterexamples here, but skip the entire test :-)?
You *could* just cut the expectations short after "happens here" in those cases, couldn't you?

ekpyron (approved):
Not sure how @axic feels about the coverage decrease ;-). And we could also do something more elaborate here IIUC... but it'd be fine with me, so tentative approval...

test/smtcallback.js
leonardoalt: Yes, I found that easier. It's only a handful of tests and the counterexamples are tested on the rest so, no harm I'd say. Just trying to keep it simple. On the other hand, I could also do that, yes.

leonardoalt: Updated

ekpyron:
Tests failing.

leonardoalt:
> Tests failing.

No way

leonardoalt:
We should also wait for https://app.circleci.com/pipelines/github/ethereum/solidity/11293/workflows/3cdf4886-1d63-42b0-b84e-f59e8ead7d6f/jobs/541919
that's the `t_ems_solcjs` from the Solidity PR using this PR

ekpyron (approved):
So yeah... still fine with me... coverage is still an open question...

leonardoalt:
Added a test directly here, let's see if that affects coverage. Otherwise, the Solidity `t_ems_solcjs` do test the added lines.

leonardoalt:
Didn't change it, I think we should merge as is.

ekpyron:
> Didn't change it, I think we should merge as is.

Since coverage is set to be required here, that'd need @axic's approval and doing then, I guess :-).

leonardoalt:
Ah, right :p

axic:
> coverage is still an open question...

Coverage has been steadily decreasing since I made the effort to reach close to 100% coverage a year or two ago 😅

axic:
Which PR introduced SMTIgnoreCex ?

leonardoalt:
@axic: https://github.com/ethereum/solidity/pull/10477 which is actually waiting for this. Right now it points to this branch, so `t_ems_solcjs` pass there

axic:
That PR only seems to be using that field, it did not introduce it?

leonardoalt:
@axic it also introduced it in `test/libsolidity/SMTCheckerTest.cpp`:
https://github.com/ethereum/solidity/pull/10477/files#r538771272

test/smtcallback.js
```diff
@@ -146,6 +151,11 @@ tape('SMTCheckerCallback', function (t) {
         continue;
       }
 
+      let ignoreCounterexamples = false;
```
axic, line 154: If you use `ignoreCex` above and in the test sources, perhaps use the same here?

test/smtcallback.js
```diff
@@ -34,7 +34,12 @@ function expectErrors (errors, expectations) {
     if (errors[i].includes('Error trying to invoke SMT solver') || expectations[i].includes('Error trying to invoke SMT solver')) {
       continue;
     }
-    expectations[i] = expectations[i].replace('happens here.', 'happens here');
+    expectations[i] = expectations[i].replace('happens here.', 'happens here').split('\\n').join('\n');
+    errors[i] = errors[i].replace('happens here.', 'happens here').split('\\n').join('\n');
+    if (ignoreCex) {
+      expectations[i] = expectations[i].split('\n')[0];
```
axic, line 40: I don't fully understand why is it done this way. Why not just cut off anything after `Counterexamples:` ?

test/smtcallback.js
leonardoalt: I wanted to, but in the `test` object it's also called `ignoreCex`, I wasn't sure about shadowing in the assignment below

test/smtcallback.js
leonardoalt: Updated

test/smtcallback.js
```diff
@@ -34,7 +34,12 @@ function expectErrors (errors, expectations) {
     if (errors[i].includes('Error trying to invoke SMT solver') || expectations[i].includes('Error trying to invoke SMT solver')) {
       continue;
     }
-    expectations[i] = expectations[i].replace('happens here.', 'happens here');
+    expectations[i] = expectations[i].replace('happens here.', 'happens here').split('\\n').join('\n');
+    errors[i] = errors[i].replace('happens here.', 'happens here').split('\\n').join('\n');
+    if (ignoreCex) {
+      expectations[i] = expectations[i].split('\nCounterexample')[0];
```
axic, line 40: If we do this, do we actually need to do that split outside of this branch?

leonardoalt: I think so. Somehow, weirdly it reads one of them as `\\n` and the other as `\n`

leonardoalt: So even if we split at `Counterexample`, `\\n` won't match `\n` when before `Counterexample`

leonardoalt: When I print `test.expectations` and `test.errors`:
```
[
  '// Warning 2529: (82-89): CHC: Empty array "pop" happens here.\\nCounterexample:\\na = []\\n\\n\\n\\nTransaction trace:\\nconstructor()\\nState: a = []\\nf()'
]
[
  'CHC: Empty array "pop" happens here.\n' +
    'Counterexample:\n' +
    'a = []\n' +
    '\n' +
    '\n' +
    '\n' +
    'Transaction trace:\n' +
    'constructor()\n' +
    'State: a = []\n' +
    'f()'
]
```

test/smtcallback.js
```diff
@@ -34,8 +34,19 @@ function expectErrors (errors, expectations) {
...
       continue;
     }
-    expectations[i] = expectations[i].replace('happens here.', 'happens here');
-    if (!errors[i].includes(expectations[i])) {
+    // Expectations containing counterexamples might have many '\n' in a single line.
+    // When these are read, they come as '\\n', whereas the actual error from the compile has '\n'.
+    // Therefore we need to replace '\\n' by '\n' in the expectations.
+    // Function `replace` only replaces the first occurrence, and `replaceAll` is not standard yet.
+    // Replace all '\\n' by '\n' via split & join.
+    expectations[i] = expectations[i].split('\\n').join('\n');
```
leonardoalt, line 42: This is needed here for the case where we actually want to compare the counterexamples

test/smtcallback.js
```diff
@@ -146,6 +157,11 @@ tape('SMTCheckerCallback', function (t) {
         continue;
       }
 
+      let ignoreCex = false;
+      if (source.includes('// SMTIgnoreCex: yes')) {
```
axic, line 161: Couldn't actually just do `let ignoreCex =  source.includes('// SMTIgnoreCex: yes')` ?

leonardoalt: Probably

test/smtcallback.js
```diff
@@ -34,8 +34,19 @@ function expectErrors (errors, expectations) {
     if (errors[i].includes('Error trying to invoke SMT solver') || expectations[i].includes('Error trying to invoke SMT solver')) {
       continue;
     }
-    expectations[i] = expectations[i].replace('happens here.', 'happens here');
-    if (!errors[i].includes(expectations[i])) {
+    // Expectations containing counterexamples might have many '\n' in a single line.
+    // When these are read, they come as '\\n', whereas the actual error from the compile has '\n'.
```
axic, line 38:
```suggestion
    // When these are read, they come as '\\n', whereas the actual error from the compiler has '\n'.
```

axic, line 38:
```suggestion
    // These are stored escaped in test format (as '\\n'), whereas the actual error from the compile has '\n'.
```

axic (approved):

PR #511, opened 2021-01-04
Use correct url for binaries.

### LINKED ISSUE #510: Forbidden response from loadRemoteVersion
Maybe the url is outdated and binaries should be retrieved from `https://binaries.soliditylang.org/bin`

chriseth:
Fixes #510

### DIFF
--- a/wrapper.js
+++ b/wrapper.js
@@ -328,7 +328,7 @@ function setupMethods (soljson) {
     // instead of from the local filesystem.
     loadRemoteVersion: function (versionString, cb) {
       var mem = new MemoryStream(null, {readable: false});
-      var url = 'https://solc-bin.ethereum.org/bin/soljson-' + versionString + '.js';
+      var url = 'https://binaries.soliditylang.org/bin/soljson-' + versionString + '.js';
       https.get(url, function (response) {
         if (response.statusCode !== 200) {
           cb(new Error('Error retrieving binary: ' + response.statusMessage));


chriseth:
Do we also want to change the `bin`-part of the path?

wrapper.js
```diff
@@ -328,7 +328,7 @@ function setupMethods (soljson) {
     // instead of from the local filesystem.
     loadRemoteVersion: function (versionString, cb) {
       var mem = new MemoryStream(null, {readable: false});
-      var url = 'https://solc-bin.ethereum.org/bin/soljson-' + versionString + '.js';
+      var url = 'https://binaries.solidity-lang.org/bin/soljson-' + versionString + '.js';
```
axic, line 331:
```suggestion
      var url = 'https://binaries.soliditylang.org/bin/soljson-' + versionString + '.js';
```

cameel: Yeah, no hyphen.

cameel:
> Do we also want to change the bin-part of the path?

Yeah, this way we'd be using the wasm rebuilds even below 0.6.2. But this requires a bit more work than just changing the URL:
- `bin/` is the only directory containing nightlies. Is `loadRemoteVersion()` supposed to allow accessing nightlies? If so, you would have to parse the version string and fall back to `bin/` if it's a nightly.
- Releases should now come from `emscripten-wasm32/` but there must be a fallback to `emscripten-asmjs/` if only an asm.js release is available.

ekpyron (approved):

chriseth:
Squashed.

ekpyron (approved):

ekpyron:
(implicit in my approval by the way is that we can still change the bin part later if we want, but changing the base url first doesn't hurt)

cameel (approved):

PR #521, opened 2021-04-06
Use SMT engine

leonardoalt:
And prepare for smtchecker pragma removal

### DIFF
--- a/test/smtcallback.js
+++ b/test/smtcallback.js
@@ -1,3 +1,4 @@
+const assert = require('assert');
 const tape = require('tape');
 const fs = require('fs');
 const path = require('path');
@@ -7,7 +8,6 @@ const smtchecker = require('../smtchecker.js');
 const smtsolver = require('../smtsolver.js');
 
 var preamble = 'pragma solidity >=0.0;\n// SPDX-License-Identifier: GPL-3.0\n';
-var pragmaSMT = 'pragma experimental SMTChecker;\n';
 
 function collectErrors (solOutput) {
   if (solOutput === undefined) {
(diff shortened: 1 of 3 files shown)


leonardoalt:
https://github.com/ethereum/solidity/pull/11202/commits/a33373095235a6e290ee957eb641debb9c4da89a
shows that this also works for upcoming versions

test/smtcallback.js
```diff
@@ -72,10 +72,20 @@ tape('SMTCheckerCallback', function (t) {
       return { error: 'Fake SMT solver error.' };
     };
 
+    var pragmaSMT = '';
+    var settings = {};
+    // `pragma experimental SMTChecker;` was deprecated in 0.8.4
+    if (!semver.gt(solc.semver(), '0.8.3')) {
```
bshastry, line 78: Perhaps `lte` for better readability instead of not greater than?

test/smtcallback.js
```diff
@@ -152,9 +162,13 @@ tape('SMTCheckerCallback', function (t) {
...
 
-      if (source.includes('// SMTEngine')) {
-        st.skip('Test requires specific SMTChecker engine.');
-        continue;
+      var engine;
+      if (source.includes('// SMTEngine:')) {
+        let re = new RegExp('// SMTEngine: (\\w+)');
+        let m = source.match(re);
+        assert(m !== undefined);
+        assert(m.length >= 2);
```
bshastry, line 170: If multiple engines are defined, they (second engine onwards) are silently ignored. Should we instead mandate single engine setting only or would that break something?

test/smtcallback.js
leonardoalt: the problem with that is pre-releases, so `< 0.8.4` actually wouldn't work for the solidity PR itself. That's why here and in other places we do this instead

test/smtcallback.js
leonardoalt: @bshastry fixed

leonardoalt: and added tests for that

test/smtcallback.js
```diff
@@ -152,9 +162,20 @@ tape('SMTCheckerCallback', function (t) {
...
+        let idx = source.indexOf(option);
+        if (source.indexOf(option, idx + 1) !== -1) {
+          st.skip('SMTEngine option given multiple times.');
+          st.end();
+          return;
+        }
+        let re = new RegExp(option + '(\\w+)');
+        let m = source.match(re);
+        assert(m !== undefined);
+        assert(m.length >= 2);
```
bshastry, line 177: If the early return for multiple SMT engines is not hit, couldn't this assert be stricter i.e., `m.length == 2`?

bshastry (approved):
Minor comment about possibly stricter assertion. Looks good otherwise, so I'm going to approve.

test/smtcallback.js
leonardoalt: I don't think so, this array actually has a bunch of info about the match itself, where the first element is the full match, second is the word we were looking for, third is index and so on.

leonardoalt:
@chriseth @axic I think the coverage decrease is due to parts of the tests already being prepared for https://github.com/ethereum/solidity/pull/11202 which is not tested yet by the tests here, but are in that PR. Can one of you merge it? I don't have access

PR #524, opened 2021-05-20
Proper SMT CHC tests

leonardoalt:
Depends on https://github.com/ethereum/solidity/pull/11421

This PR adds *proper* SMTChecker tests. Currently, the SMTChecker tests are actually broken, many tests are ignored, and the smtlib version of the SMTCheckre encoding is not being tests at all.

The main `smtCheckerTests` in this code are mainly used by the Solidity `t_ems_solcjs` job, which copies the `smtCheckerTests` into the `solc-js` tests and runs `test/smtcallback.js` from this repo. It now requires `soljson` version `>=0.7.2` because of the warning format.

If `soljson < 0.8.5`, it runs the static `z3` inside `soljson` and compares that to the test expectation.
Else, it runs /\ plus a local `z3` binary using Solidity's `smtlib2` encoding, and compares all of them.
The comparison now is much more precise than before: it compares each output location, which represents a property.

Note that safe properties are not reported, so if one solver reports a location and another doesn't, there's an inconsistency. The exception to this is when this inconsistency happens for the BMC engine, where one solver probably solved it safe with CHC, and another solver couldn't prove it safe with CHC, and reports an unsafe false positive via BMC.

The test only fails if there's a real inconsistency, as opposed to the stronger `soltest`. I think that's fair since the JS/smtlib2 environment might give more nondeterministic answers compared to the embedded C++ z3.

### DIFF
--- a/smtsolver.js
+++ b/smtsolver.js
@@ -3,12 +3,13 @@ var execSync = require('child_process').execSync;
 var fs = require('fs');
 var tmp = require('tmp');
 
+// Timeout in ms.
 const timeout = 10000;
 
 var potentialSolvers = [
   {
     name: 'z3',
-    params: '-smt2 rlimit=20000000 rewriter.pull_cheap_ite=true fp.spacer.q3.use_qgen=true fp.spacer.mbqi=false fp.spacer.ground_pobs=false'
+    params: '-smt2 timeout=' + timeout + ' rewriter.pull_cheap_ite=true fp.spacer.q3.use_qgen=true fp.spacer.mbqi=false fp.spacer.ground_pobs=false'
   },
   {
     name: 'cvc4',
--- a/smtsolver.js
+++ b/smtsolver.js
@@ -45,7 +45,9 @@ function solve (query) {
     if (
       !solverOutput.startsWith('sat') &&
       !solverOutput.startsWith('unsat') &&
-      !solverOutput.startsWith('unknown')
+      !solverOutput.startsWith('unknown') &&
+      !solverOutput.startsWith('(error') &&
+      !solverOutput.startsWith('error')
     ) {
       throw new Error('Failed to solve SMT query. ' + e.toString());
     }
--- a/smtsolver.js
+++ b/smtsolver.js
@@ -4,29 +4,28 @@ var fs = require('fs');
 var tmp = require('tmp');
 
 // Timeout in seconds.
-const timeout = 1000;
+const timeout = 300;
 
 var potentialSolvers = [
   {
     name: 'Eldarica Vanilla',
     command: 'eld',
-    params: '-horn -t:' + timeout
+    params: '-horn'
   },
   {
     name: 'Eldarica No Abstraction',
     command: 'eld',
-    params: '-horn -ssol -t:' + timeout + ' -abstract:off'
-  }
-  /*
+    params: '-horn -abstract:off'
+  },
   {
     name: 'Eldarica Term Abstraction',
     command: 'eld',
-    params: '-horn -t:' + timeout + ' -abstract:term'
+    params: '-horn -abstract:term'
   },
   {
     name: 'Eldarica Oct Abstraction',
     command: 'eld',
-    params: '-horn -t:' + timeout + ' -abstract:oct'
+    params: '-horn -abstract:oct'
   },
   {
     name: 'Spacer Vanilla',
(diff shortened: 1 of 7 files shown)


test/smtcallback.js
```diff
@@ -7,88 +7,198 @@ const solc = require('../index.js');
...
-    errors.push(error.message);
   }
-  return errors;
+  return true;
 }
 
-function expectErrors (expectations, errors, ignoreCex) {
-  if (errors.length !== expectations.length) {
-    return false;
+function collectFiles (testdir) {
```
leonardoalt, line 85: Just extracted from the main loop

test/smtcallback.js
```diff
@@ -7,88 +7,198 @@ const solc = require('../index.js');
...
+        dirs.push(file);
+      } else {
+        sources.push(file);
+      }
+    }
   }
+  return sources;
+}
+
+function createTests (sources, st) {
```
leonardoalt, line 104: Extracted from the main loop

test/smtcallback.js
```diff
@@ -7,88 +7,198 @@ const solc = require('../index.js');
...
+  }
+  return tests;
+}
+
+function collectStringErrors (solOutput) {
+  assert(solOutput !== undefined);
+  return solOutput.errors.filter(error => !error.message.includes('This is a pre-release')).map(error => error.message);
+}
+
+function expectStringErrors (a, b) {
```
leonardoalt, line 152: These two functions have the same behavior as before and are only used by the first test.

test/smtcallback.js
```diff
@@ -138,82 +255,27 @@ tape('SMTCheckerCallback', function (t) {
...
       // `pragma experimental SMTChecker;` was deprecated in 0.8.4
       if (semver.gt(solc.semver(), '0.8.3')) {
-        let engine = test.engine !== undefined ? test.engine : 'all';
+        const engine = test.engine !== undefined ? test.engine : 'all';
         settings = { modelChecker: { engine: engine } };
+
+        // 0.8.5 introduced the `solvers` option.
+        // We force the compiler to use smtlib2 in this test.
+        if (semver.gt(solc.semver(), '0.8.4')) {
+          settings.modelChecker.solvers = ['smtlib2'];
```
leonardoalt, line 275: Because of this, from 0.8.5 on we *don't* use the static z3 inside soljson.js for this test anymore, but require a z3 binary in the system.

leonardoalt:
I really don't understand this coverage tool.

cameel:
@leonardoalt Is this still work in progress or is it just waiting to be reviewed?

leonardoalt:
Waiting to be reviewed

smtsolver.js
```diff
@@ -3,12 +3,13 @@ var execSync = require('child_process').execSync;
...
 
-const timeout = 10000;
+// Timeout in ms.
+const timeout = 1000;
 
 var potentialSolvers = [
   {
     name: 'z3',
-    params: '-smt2 rlimit=20000000 rewriter.pull_cheap_ite=true fp.spacer.q3.use_qgen=true fp.spacer.mbqi=false fp.spacer.ground_pobs=false'
+    params: '-smt2 timeout=' + timeout + ' rewriter.pull_cheap_ite=true fp.spacer.q3.use_qgen=true fp.spacer.mbqi=false fp.spacer.ground_pobs=false'
```
axic, line 12: ~~Why time limit and not resource limit?~~

Ah this is just so that both cvc4/z3 have the same limit.

smtsolver.js
```diff
@@ -44,7 +45,9 @@ function solve (query) {
     if (
       !solverOutput.startsWith('sat') &&
       !solverOutput.startsWith('unsat') &&
-      !solverOutput.startsWith('unknown')
+      !solverOutput.startsWith('unknown') &&
+      !solverOutput.startsWith('(error') &&
```
axic, line 49: What outputs `(`?

axic: @leonardoalt ^^

leonardoalt: ah, Eldarica outputs errors in some sort of smtlib sexpr, like `(error something failed lalala)` which is caught here

axic:
I suppose #565 needs to be merged first?

leonardoalt:
Ah yes, this branch is all over the place. Will need to rethink it after the other one gets merged

leonardoalt:
Closing cus need to rewrite it anyway.

PR #533, opened 2021-07-09
[solcjs] Base path stripping matching `solc`

cameel:
This is the solc-js part of https://github.com/ethereum/solidity/issues/11408.

The PR changes the base path stripping logic to more or less match https://github.com/ethereum/solidity/pull/11617 / https://github.com/ethereum/solidity/pull/11545. This is a simplified version so I did not try to get all corner cases right and test coverage is minimal (I can't test it well without access to the JSON input anyway).

I only tested it on Linux. Hopefully CI runs on Windows too.

One extra change is that the PR enables import callback even without `--base-path` option because this is how it works on `solc`. I did not enable it for `--standard-json` because we do not have `--allow-paths` here and therefore the callback could kick in even without the user realizing it.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -4,6 +4,7 @@
 var originalUncaughtExceptionListeners = process.listeners("uncaughtException");
 
 var fs = require('fs-extra');
+var os = require('os');
 var path = require('path');
 var solc = require('./index.js');
 var smtchecker = require('./smtchecker.js');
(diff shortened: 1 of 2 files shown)


leonardoalt (commented):
LGTM, but maybe @axic should take a second look. It's not long!

PR #534, opened 2021-07-09
[solcjs] `--pretty-json` and `--verbose` options

cameel:
Somewhat related to https://github.com/ethereum/solidity/issues/11583.

This PR adds two options
- `--pretty-print` - we already have that in solc and soon it will work for Standard jSON too (https://github.com/ethereum/solidity/pull/11639).
- `--verbose` - prints extra info, most importantly the Standard JSON input to the compiler. It's most useful when not using `--standard-json` option because then it's the script that constructs the input JSON. I wanted to see it pretty much every time I was solving a problem with solc-js and I always ended up hacking it in with debug prints. Having this as an option would make working with solc-js much more convenient for me and I think it might be useful for others too.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -38,7 +38,8 @@ program
   .option('--abi', 'ABI of the contracts.')
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
   .option('--base-path <path>', 'Automatically resolve all imports inside the given path.')
-  .option('-o, --output-dir <output-directory>', 'Output directory for the contracts.');
+  .option('-o, --output-dir <output-directory>', 'Output directory for the contracts.')
+  .option('-p, --pretty-json', 'Pretty-print all JSON output.', false);
 
 program.parse(process.argv);
 const options = program.opts();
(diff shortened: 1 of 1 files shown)


solcjs
```diff
@@ -52,18 +54,30 @@ function stripBasePath(path) {
...
+    console.log('>>> Compiling:\n' + reformatJsonIfRequested(input) + "\n")
+  var output = reformatJsonIfRequested(solc.compile(input, callbacks));
 
   try {
     var inputJSON = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output), smtsolver.smtSolver);
     if (inputJSON) {
-      output = solc.compile(JSON.stringify(inputJSON), callbacks);
+      if (program.verbose)
+        console.log('>>> Retrying compilation with SMT:\n' + toFormattedJson(inputJSON) + "\n")
+      output = reformatJsonIfRequested(solc.compile(JSON.stringify(inputJSON), callbacks));
```
cameel, line 80: @leonardoalt This is a bit off-topic in this PR, but why do we have to retry here? Can't it be detected based on input alone?

axic (approved):
The goal and changes look ok to me (except the question to @leonardoalt), but I am a bit wary of adding too many features here, as that also means maintaining them.

I'd also suggest to ensure the options which are available match 100% with solc.

cameel:
These two are actually worth it in my opinion. Maintenance cost is minimal and they save time when debugging solc-js.

> I'd also suggest to ensure the options which are available match 100% with solc.

They're unfortunately not but I think they're close enough.
- `--verbose` was not needed so far in `solc`. At least in the sense used here - it does not construct JSON input but gets it as a parameter.
- `--pretty-json` is present in `solc` but it has `--json-indent` which I did not add here to keep the interface minimal. Instead I just hard-coded  set the default indent here to 4 spaces.

axic:
> `--pretty-json`.is present in `solc` but it has `--json-indent` which I did not add here to keep the interface minimal. Instead I just hard-coded set the default indent here to 4 spaces.

I guess as long as the behaviour is not different (just lacks features) it is fine.

PR #540, opened 2021-08-12
Use prepublishOnly.

chriseth:


chriseth:
Seems we actually need the old behaviour.

ekpyron (approved):
I didn't actually read up on this myself, but if we're a bit conscious about it at the next release, I'd say we can merge this.

package.json
```diff
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "lint": "node ./node_modules/semistandard/bin/cmd.js",
-    "prepublish": "node downloadCurrentVersion.js && node verifyVersion.js",
-    "pretest": "npm run lint",
+    "prepublishOnly": "node downloadCurrentVersion.js && node verifyVersion.js",
+    "pretest": "npm run lint && npm run prepublishOnly",
```
ekpyron, line 12: This now tries to run ``verifyVersion.js`` before our ``t_ems_solcjs`` tests triggered from the solidity repo with CI build soljson.js - which fails the version test...

PR #544, opened 2021-08-27
[solcjs] `--include-path` option

cameel:
Implements the `solc-js` part of https://github.com/ethereum/solidity/pull/11848.
Depends on #547.

The implementation is simplified compared to `solc` and argument processing from the `commander` works a bit differently than in `boost::program_options`:
- There's no check against path conflicts (two files specified on the CLI getting the same source unit name after base path/include path stripping).
- There's no check against ambiguous imports (one import matching multiple files in base path/include paths).
- `commander` allows specifying multiple paths while using the option only once `--include-path path1/ path2/ path3/`. This is a bit inconvenient because if it's the last option, it consumes the names of the input files. E.g. `solcjs --base-path base/ --include-paths node_modules/ MyContract.sol` will treat `MyContract.sol` as one of the include paths. The library does not allow disabling this behavior.
    - The boost syntax with option used multiple times works too: `--include-path path1/ --include-path  path2/ --include-path  path3/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,17 @@ solcjs --help
 
 To compile a contract that imports other contracts via relative paths:
 ```bash
-solcjs --bin --base-path . ./MainContract.sol
+solcjs --bin --include-path node_modules/ --base-path . MainContract.sol
 ```
-The option ``--base-path`` enables automatic loading of imports from the filesystem and
-takes a path as argument that contains the source files.
+Use the ``--base-path`` and ``--include-path`` options to describe the layout of your project.
+``--base-path`` represents the root of your own source tree while ``--include-path`` allows you to
+specify extra locations containing external code (e.g. libraries installed with a package manager).
+
+Note: ensure that all the files you specify on the command line are located inside the base path or
+one of the include paths.
+The compiler refers to files from outside of these directories using absolute paths.
+Having absolute paths in contract metadata will result in your bytecode being reproducible only
+when it's placed in these exact absolute locations.
 
 Note: this commandline interface is not compatible with `solc` provided by the Solidity compiler package and thus cannot be
 used in combination with an Ethereum client via the `eth.compile.solidity()` RPC method. Please refer to the
--- a/solcjs
+++ b/solcjs
@@ -37,7 +37,13 @@ program
   .option('--bin', 'Binary of the contracts in hex.')
   .option('--abi', 'ABI of the contracts.')
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
-  .option('--base-path <path>', 'Automatically resolve all imports inside the given path.')
+  .option('--base-path <path>', 'Root of the project source tree. ' +
+    'The import callback will attempt to interpret all import paths as relative to this directory.'
+  )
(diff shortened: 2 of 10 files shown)


cameel:
This PR now depends on #547.

README.md
```diff
@@ -28,10 +28,17 @@ solcjs --help
 
 To compile a contract that imports other contracts via relative paths:
 ```bash
-solcjs --bin --base-path . ./MainContract.sol
+solcjs --bin --include-path node_modules/ --base-path . MainContract.sol
 ```
-The option ``--base-path`` enables automatic loading of imports from the filesystem and
-takes a path as argument that contains the source files.
+Use the ``--base-path`` and `--include-path`` options to describe the layout of our project.
```
leonardoalt, line 33:
```suggestion
Use the ``--base-path`` and ``--include-path`` options to describe the layout of your project.
```

README.md
```diff
@@ -28,10 +28,17 @@ solcjs --help
...
+``--base-path`` represents root of your own source tree while ``--include-path`` allows you to
```
leonardoalt, line 34:
```suggestion
``--base-path`` represents the root of your own source tree while ``--include-path`` allows you to
```

solcjs
```diff
@@ -37,16 +43,19 @@ function abort (msg) {
...
+      try {
+        return {'contents': fs.readFileSync(prefixedSourcePath).toString('utf8')}
+      } catch (e) {
+        return {error: 'Error reading ' + prefixedSourcePath + ': ' + e};
+      }
     }
-  } else
-    return { error: 'File not found at ' + sourcePath}
+  }
+  return {error: 'File not found inside the base path or any of the include paths.'}
```
leonardoalt, line 58: Maybe still mention which source could not be found?

solcjs
```diff
@@ -37,16 +43,19 @@ function abort (msg) {
...
 
 function readFileCallback(sourcePath) {
-  if (options.basePath)
-    sourcePath = options.basePath + '/' + sourcePath;
-  if (fs.existsSync(sourcePath)) {
-    try {
-      return { 'contents': fs.readFileSync(sourcePath).toString('utf8') }
-    } catch (e) {
-      return { error: 'Error reading ' + sourcePath + ': ' + e };
+  const prefixes = [options.basePath ? options.basePath : ""].concat(options.includePath)
```
leonardoalt, line 46: Does anything guarantee that `options.includePath` is an array?

leonardoalt (approved):
wait why did I approve

README.md
cameel: Haha, interesting that `our` is just a typo away from `your` and results in something that still kinda makes sense :)

solcjs
cameel: It's redundant. The part you see here is just a fragment of the message that the callback returns to the compiler. The compiler will prepend a message that already includes the source unit name of the file, which is why I removed it.

I do include it in other messages above because there I know the full path (i.e. source unit name with a base path or an include path prepended) so it actually adds some useful information.

chriseth (approved):

PR #546, opened 2021-09-04
Adding parameter to solcjs for the optimizer run option

rorye:
Adding parameter to specify the optimizer runs options. Example use: 

`node ./solcjs --optimize-runs 666 --bin test/resources/fixtureSmoke.sol`

Resolves https://github.com/ethereum/solidity/issues/11857

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -15,16 +15,31 @@ process.removeAllListeners('uncaughtException');
 var commander = require('commander');
 
 const program = new commander.Command();
+const commanderParseInt = function(value) {
+  const parsedValue = parseInt(value, 10);
+  if (isNaN(parsedValue)) {
+    throw new commander.InvalidArgumentError("Not a valid integer.");
+  }
+  return parsedValue;
+};
+
 program.name('solcjs');
 program.version(solc.version());
 program
   .option('--version', 'Show version and exit.')
   .option('--optimize', 'Enable bytecode optimizer.', false)
+  .option(
+    '--optimize-runs <optimize-runs>',
+    'The number of runs specifies roughly how often each opcode of the deployed code will be executed across the lifetime of the contract. ' +
+    'Lower values will optimize more for initial deployment cost, higher values will optimize more for high-frequency usage.',
+    commanderParseInt
+  )
   .option('--bin', 'Binary of the contracts in hex.')
   .option('--abi', 'ABI of the contracts.')
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
   .option('--base-path <path>', 'Automatically resolve all imports inside the given path.')
   .option('-o, --output-dir <output-directory>', 'Output directory for the contracts.');
+
 program.parse(process.argv);
 const options = program.opts();
 
@@ -131,7 +146,8 @@ var output = JSON.parse(solc.compile(JSON.stringify({
   language: 'Solidity',
   settings: {
     optimizer: {
-      enabled: options.optimize
+      enabled: options.optimize,
+      runs: options.optimizeRuns,
     },
     outputSelection: {
       '*': {
--- a/test/cli.js
+++ b/test/cli.js
@@ -37,6 +37,19 @@ tape('CLI', function (t) {
     spt.end();
   });
 
+  t.test('--bin --optimize-runs 666', function (st) {
+    var spt = spawn(st, './solcjs --bin --optimize-runs 666 test/resources/fixtureSmoke.sol');
+    spt.stderr.empty();
+    spt.succeeds();
+    spt.end();
+  });
+
+  t.test('--bin --optimize-runs not-a-number', function (st) {
+    var spt = spawn(st, './solcjs --bin --optimize-runs not-a-number test/resources/fixtureSmoke.sol');
+    spt.stderr.match(/^error: option '--optimize-runs <optimize-runs>' argument 'not-a-number' is invalid/);
+    spt.end();
+  });
+
   t.test('invalid file specified', function (st) {
     var spt = spawn(st, './solcjs --bin test/fileNotFound.sol');
     spt.stderr.match(/^Error reading /);


solcjs
```diff
@@ -14,12 +14,21 @@ var smtsolver = require('./smtsolver.js');
...
+  return parsedValue;
+};
+
 var program = new commander.Command();
 program.name('solcjs');
 program.version(solc.version());
 program
   .option('--version', 'Show version and exit.')
   .option('--optimize', 'Enable bytecode optimizer.', false)
+  .option('--optimize-runs <optimize-runs>', 'The number of runs specifies roughly how often each opcode of the deployed code will be executed across the life-time of the contract.', parseOptimizeRuns)
```
cameel, line 31: This line is pretty long. I'd wrap it.
I'd also add part of the description that `solc` uses:
```suggestion
  .option(
    '--optimize-runs <optimize-runs>',
    'The number of runs specifies roughly how often each opcode of the deployed code will be executed across the lifetime of the contract. ' +
    'Lower values will optimize more for initial deployment cost, higher values will optimize more for high-frequency usage.',
    parseOptimizeRuns
  )
```

solcjs
```diff
@@ -130,7 +139,8 @@ var output = JSON.parse(solc.compile(JSON.stringify({
   language: 'Solidity',
   settings: {
     optimizer: {
-      enabled: program.optimize
+      enabled: program.optimize,
+      runs: program.optimizeRuns
```
cameel, line 143:
```suggestion
      runs: program.optimizeRuns,
```

cameel (commented):
Needs some cosmetic tweaks. Other than that it looks good.

cameel (approved):

solcjs
```diff
@@ -14,12 +14,26 @@ var smtsolver = require('./smtsolver.js');
...
+};
+
 var program = new commander.Command();
 program.name('solcjs');
 program.version(solc.version());
 program
   .option('--version', 'Show version and exit.')
   .option('--optimize', 'Enable bytecode optimizer.', false)
+  .option(
+    '--optimize-runs <optimize-runs>',
```
axic, line 32: It would be nice if this description here matches the main `solc` CLI:
>   --optimize-runs n (=200)
>                       Set for how many contract runs to optimize. Lower values
>                       will optimize more for initial deployment cost, higher 
>                       values will optimize more for high-frequency usage.

axic: I think the version in this PR is nicer, so I would suggest to create a PR against https://github.com/ethereum/solidity to update the other tool as well.

solcjs
```diff
@@ -14,12 +14,26 @@ var smtsolver = require('./smtsolver.js');
 process.removeAllListeners('uncaughtException');
 var commander = require('commander');
 
+const parseOptimizeRuns = function(value) {
+  const parsedValue = parseInt(value, 10);
+  if (isNaN(parsedValue)) {
+    return value;
```
axic, line 20: The official documentation suggests to raise an error here:
```suggestion
    throw new commander.InvalidArgumentError('Not a number.');
```

solcjs
cameel: Yeah, I thought that too, which is why I only suggested adding part of the description from solc.

Syncing it the other way seems like a good idea.

solcjs
cameel: I tested it locally and I think it's fine. It does report errors for non-int values properly. If the value is not an int, it just puts it into JSON and lets it do any further validation.

Actually, one thing I'm not certain about is if it matches solc's full integer range. In JS there are only floats so some larger ints that the compiler would accept might be passed to it as strings. Or worse - be rounded? @rorye, I think we should have a test for the largest possible int value.

> why does it work with the current code (it checks that the type of the input and the output differs?)

I don't understand this question. AFAIK commander does not check the type at all, which is why it's being converted to int here.

axic: The code above does not exceptions just returns the input, yet [this test case](https://github.com/ethereum/solc-js/pull/546/files#diff-aae216853526bb5eb76ffeff66cc7f56184296adde901fd9d48c08e925723a8aR49) seemingly works. I'm not sure how.

cameel: It works because it's `compile()` that raises the exception. The code here just returns non-int values unchanged which makes the compiler binary report an error later.

axic: Ah okay, I guess it is fine using that but I'd document it. Alternatively just do the `throw` as I suggested, that is in line with the CLI library.

rorye: @axic - Added the throw as suggested and the corresponding catch. I could not use `commander.InvalidArgumentError` as it resulted in 

```
Error: "commander.InvalidArgumentError is not a constructor
```
Perhaps the version of commander we are using is a little old.

cameel: @rorye It is old. I needed to upgrade it for my own PR too. I have extracted the change that upgrades it into a separate PR (#547). You can rebase yours on it.

solcjs
```diff
@@ -14,18 +14,37 @@ var smtsolver = require('./smtsolver.js');
...
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
   .option('--base-path <path>', 'Automatically resolve all imports inside the given path.')
   .option('-o, --output-dir <output-directory>', 'Output directory for the contracts.');
-program.parse(process.argv);
+
+try {
+  program.parse(process.argv);
+} catch(e) {
+  abort(e.message);
+}
```
cameel, line 47: This should only catch validation errors. Unexpected exceptions should cause a crash.

rorye: Thanks, have updated accordingly.

cameel: @rorye Can you rebase on `master` and squash review fixes?

rorye: @cameel please see update.

solcjs
```diff
@@ -15,17 +15,39 @@ process.removeAllListeners('uncaughtException');
...
   .option('--bin', 'Binary of the contracts in hex.')
   .option('--abi', 'ABI of the contracts.')
   .option('--standard-json', 'Turn on Standard JSON Input / Output mode.')
   .option('--base-path <path>', 'Automatically resolve all imports inside the given path.')
   .option('-o, --output-dir <output-directory>', 'Output directory for the contracts.');
-program.parse(process.argv);
+try {
+  program.parse(process.argv);
+} catch(e) {
+  if(e instanceof commander.InvalidArgumentError) {
```
axic, line 45: Is all this really needed? According to the [documentation](https://www.npmjs.com/package/commander#override-exit-and-output-handling) it should just abort itself.

solcjs
```diff
@@ -15,17 +15,39 @@ process.removeAllListeners('uncaughtException');
 var commander = require('commander');
 
 const program = new commander.Command();
+const parseOptimizeRuns = function(value) {
+  const parsedValue = parseInt(value, 10);
+  if (isNaN(parsedValue)) {
+    throw new commander.InvalidArgumentError('The "runs" setting must be an unsigned number');
```
axic, line 21: I wonder if this function should be just a generic `commanderParseInt` function?

cameel: Yeah. That's a good idea. This could be a generic function for validating unsigned ints. The message here does not even have to include the name of the argument or value because commander will prepend that to the message it displays.

solcjs
rorye: Updated

solcjs
rorye: Updated

solcjs
```diff
@@ -32,22 +32,15 @@ program
     '--optimize-runs <optimize-runs>',
     'The number of runs specifies roughly how often each opcode of the deployed code will be executed across the lifetime of the contract. ' +
     'Lower values will optimize more for initial deployment cost, higher values will optimize more for high-frequency usage.',
-    parseOptimizeRuns
+      commanderParseInt
```
cameel, line 35:
```suggestion
    commanderParseInt
```

cameel (approved):
There's just one tiny indentation issue. Also, commits need to be squashed again (we usually squash any review fixes into the original commit so that each commit makes sense on its own).

Other than that, I think it's done. Do you approve @axic?

solcjs
```diff
@@ -15,16 +15,31 @@ process.removeAllListeners('uncaughtException');
 var commander = require('commander');
 
 const program = new commander.Command();
+const commanderParseInt = function(value) {
+  const parsedValue = parseInt(value, 10);
+  if (isNaN(parsedValue)) {
+    throw new commander.InvalidArgumentError();
```
axic, line 21: Nit: I would display a simple message here, but its is fine either way.
```suggestion
    throw new commander.InvalidArgumentError("Not a valid integer.");
```

axic (approved):

cameel:
I did some minor cleanup (squashed commits, fixed whitespace) and I'm going to merge it.

At least if it passes looks like last time `coveralls` failed with a HTTP error.

solcjs
```diff
@@ -15,16 +15,31 @@ process.removeAllListeners('uncaughtException');
 var commander = require('commander');
 
 const program = new commander.Command();
+const commanderParseInt = function(value) {
+  const parsedValue = parseInt(value, 10);
```
cameel, line 19: Looks like [`parseInt()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) is **very** lax when parsing integers:
> If `parseInt` encounters a character that is not a numeral in the specified `radix`, it ignores it and all succeeding characters and returns the integer value parsed up to that point. `parseInt` truncates numbers to integer values. Leading and trailing spaces are allowed.

So something like this
```bash
./solcjs /tmp/test.sol --bin --optimize-runs 6Y7Z8___sdfas
```
is accepted and the value is `6`.

I think we would have been better off with the original version that just passed the value to solc directly. But I won't be changing this back now. I'd like to get this finally merged.

test/cli.js
```diff
@@ -37,6 +37,19 @@ tape('CLI', function (t) {
...
+  t.test('--bin --optimize-runs 666', function (st) {
+    var spt = spawn(st, './solcjs --bin --optimize-runs 666 test/resources/fixtureSmoke.sol');
+    spt.stderr.empty();
+    spt.succeeds();
+    spt.end();
+  });
+
+  t.test('--bin --optimize-runs not-a-number', function (st) {
+    var spt = spawn(st, './solcjs --bin --optimize-runs not-a-number test/resources/fixtureSmoke.sol');
+    spt.stderr.match(/^error: option '--optimize-runs <optimize-runs>' argument 'not-a-number' is invalid/);
```
axic, line 49: Aha, I see this should have had a `$/` at the end so we fully match the message. I assume it should have `Not a valid integer.` at the end.

PR #547, opened 2021-09-09
Upgrade `commander` dependency to the latest version

cameel:
Extracted from #544 because it looks useful for #546 as well.

- Newer versions of the library allow having parameters that take multiple values, which is needed in #544.
- Apparently the `InvalidArgumentError` is only available in never versions (https://github.com/ethereum/solc-js/pull/546#discussion_r705200874).

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
     "command-exists": "^1.2.8",
-    "commander": "3.0.2",
+    "commander": "^8.1.0",
     "follow-redirects": "^1.12.1",
     "fs-extra": "^0.30.0",
     "js-sha3": "0.8.0",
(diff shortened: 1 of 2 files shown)


petrosDemetrakopoulos (approved):

cameel:
@petrosDemetrakopoulos Any comments on the PR? I just want to make sure you actually looked at it and you're not a bot. We're getting random comments and "reviews" from bots pretty often :)

petrosDemetrakopoulos:
> @petrosDemetrakopoulos Any comments on the PR? I just want to make sure you actually looked at it and you're not a bot. We're getting random comments and "reviews" from bots pretty often :)

Yes, I did! I also pulled and ran tests locally. Everything seems ok.

cameel:
Great. Thanks! I'd still wait a bit for a final approval from @axic but the change is straightforward enough that I think it should not break anything.

axic (approved):

Issue #211, opened 2018-04-16
Move CI to circleci


axic:
Currently solc-js is using travis to run tests:
- `npm run test`
- `npm run coveralls`

It runs these on all stable node.js releases: 4, 6, 8 (should also do on nodejs 10)

The goal is to write a circleci configuration doing the same.

fanatid:
Out of interest, which benefits CircleCI provide (compare to travis)?

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 125.0 DAI (125.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/solc-js/211/1114).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/solc-js/211/1114).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $20,475.83 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


1. [bakaoh](https://gitcoin.co/profile/bakaoh) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/solc-js/211/1114?mutate_worker_action=approve&worker=bakaoh) | [reject worker](https://gitcoin.co/issue/ethereum/solc-js/211/1114?mutate_worker_action=reject&worker=bakaoh))_. 
    
    It's seem simple. I will get it done today.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/solc-js/211/1114).

rsercano:
@axic if you approve, please let me know so that I can start to work on this

vs77bb:
Approved @bakoah, who we've seen around here and was first on the scene. @rsercano / @rohithzr, please do keep an eye on the [Issue Explorer](gitcoin.co/explorer) for more issues like this or make a [Gitcoin Request](gitcoin.co/requests) if there are other open source projects where you'd like to help!

[...]

PR #549, opened 2021-09-15
Drop support for Travis

axic:
Travis has not been used for a long while (~2 years, see #211) and considering the recent security breach we may not want to use it again.


cameel (approved):
You should also remove the Travis badge from README.

axic:
@cameel done

cameel (approved):

PR #563, opened 2021-11-10
Sanitize nightly version string which had leading zeroes

### LINKED ISSUE #562: Older nightlies with leading zeros in version string crash `soljson`
Originally reported in https://github.com/ethereum/solc-bin/pull/21#discussion_r746961386.

`soljson-v0.4.1-nightly.2016.9.9+commit.79867f49.js` crashes solc-js because the version it reports apparently does not conform to semver. solc-js should be able to handle that more gracefully.

This happens for `--version` and `--bin`. Probably in other cases too.

### Repro
```bash
npm install solc
curl https://binaries.soliditylang.org/bin/soljson-v0.4.1-nightly.2016.9.9+commit.79867f49.js --location --output node_modules/solc
npx solcjs --version
```

Running the above produces the following exception:
```
TypeError: Invalid Version: 0.4.1-nightly.2016.09.09+commit.79867f49.Emscripten.clang
    at new SemVer (/tmp/s/node_modules/semver/semver.js:323:11)
    at compare (/tmp/s/node_modules/semver/semver.js:614:10)
    at Function.gt (/tmp/s/node_modules/semver/semver.js:643:10)
    at setupMethods (/tmp/s/node_modules/solc/wrapper.js:20:27)
    at Object.<anonymous> (/tmp/s/node_modules/solc/index.js:3:18)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
```

axic:
Fixes #562.

### DIFF
--- a/test/translate.ts
+++ b/test/translate.ts
@@ -18,6 +18,11 @@ tape('Version string to Semver translator', function (t) {
     st.equal(versionToSemver('0.4.20-nightly.2018.2.13+commit.27ef9794.Emscripten.clang'), '0.4.20-nightly.2018.2.13+commit.27ef9794.Emscripten.clang');
     st.end();
   });
+  t.test('Broken nightly with leading zeroes', function (st) {
+    st.equal(versionToSemver('0.3.6-nightly.2016.08.27+commit.91d4fa47.Emscripten.clang'), '0.3.6-nightly.2016.8.27+commit.91d4fa47.Emscripten.clang');
+    st.equal(versionToSemver('0.4.1-nightly.2016.09.09+commit.79867f49.Emscripten.clang'), '0.4.1-nightly.2016.9.9+commit.79867f49.Emscripten.clang');
+    st.end();
+  });
   t.test('Old style 0.1.1', function (st) {
     st.equal(versionToSemver('0.1.1-6ff4cd6b/RelWithDebInfo-Emscripten/clang/int'), '0.1.1+commit.6ff4cd6b');
     st.end();
@@ -35,6 +40,10 @@ tape('Version string to Semver translator', function (t) {
       versionToSemver('0.2.0-e7098958/.-Emscripten/clang/int linked to libethereum-1.1.1-bbb80ab0/.-Emscripten/clang/int'),
       '0.2.0+commit.e7098958'
     );
+    st.end();
+  });
+  t.test('Broken 0.3.4 nightly', function (st) {
+    st.equal(versionToSemver('0.3.4-0/Release-Emscripten/clang/Interpreter'), '0.3.4-nightly');
     st.end();
   });
   t.test('Old style 0.3.5', function (st) {
--- a/test/translate.ts
+++ b/test/translate.ts
@@ -21,6 +21,7 @@ tape('Version string to Semver translator', function (t) {
   t.test('Broken nightly with leading zeroes', function (st) {
     st.equal(versionToSemver('0.3.6-nightly.2016.08.27+commit.91d4fa47.Emscripten.clang'), '0.3.6-nightly.2016.8.27+commit.91d4fa47.Emscripten.clang');
     st.equal(versionToSemver('0.4.1-nightly.2016.09.09+commit.79867f49.Emscripten.clang'), '0.4.1-nightly.2016.9.9+commit.79867f49.Emscripten.clang');
+    st.equal(versionToSemver('0.4.1-nightly.2016.12.06+commit.79867f49.Emscripten.clang'), '0.4.1-nightly.2016.12.6+commit.79867f49.Emscripten.clang');
     st.end();
   });
   t.test('Old style 0.1.1', function (st) {
--- a/translate.ts
+++ b/translate.ts
@@ -1,15 +1,20 @@
 import linker from './linker';
 
 /// Translate old style version numbers to semver.
-/// Old style: 0.3.6-3fc68da5/Release-Emscripten/clang
+/// Old style: 0.4.1-nightly.2016.09.09+commit.79867f49.Emscripten.clang
+///            0.3.6-nightly.2016.08.27+commit.91d4fa47.Emscripten.clang
+///            0.3.6-3fc68da5/Release-Emscripten/clang
 ///            0.3.5-371690f0/Release-Emscripten/clang/Interpreter
 ///            0.3.5-0/Release-Emscripten/clang/Interpreter
+///            0.3.4-0/Release-Emscripten/clang/Interpreter
 ///            0.2.0-e7098958/.-Emscripten/clang/int linked to libethereum-1.1.1-bbb80ab0/.-Emscripten/clang/int
 ///            0.1.3-0/.-/clang/int linked to libethereum-0.9.92-0/.-/clang/int
 ///            0.1.2-5c3bfd4b*/.-/clang/int
 ///            0.1.1-6ff4cd6b/RelWithDebInfo-Emscripten/clang/int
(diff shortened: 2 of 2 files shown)


axic:
Needs to be squashed, but the Github editor doesn't let me do that.

translate.js
```diff
@@ -21,6 +24,11 @@ function versionToSemver (version) {
   if (version.indexOf('0.3.5-0') !== -1) {
     return '0.3.5';
   }
+  // This parses the obsolete nightly style where the date can have leading zeroes.
+  var nightly_parsed = version.match(/^([0-9]+\.[0-9]+\.[0-9]+)-nightly\.([0-9]+)\.0([1-9])\.0([1-9])(.*)$/);
```
cameel: This does work for this one nightly but technically non-zero could also happen on the first position in the day or month.

I have just pushed https://github.com/ethereum/solc-bin/pull/103 and we might actually have some nightlies like that in there.

axic: Hm? This regex checks zeroes for months/days.

axic: Ah you mean **or**. Yes, you're correct, need to improve it for that.

axic: Could you run this on old nightlies? If there is no such instance, I think keeping this simple is better.

cameel: Bad news. We have several 0.3.6 nightlies that need this:
| File name                                            | Reported version                                                 |
|------------------------------------------------------|------------------------------------------------------------------|
| `soljson-v0.3.6-nightly.2016.8.27+commit.91d4fa47.js`| `0.3.6-nightly.2016.08.27+commit.91d4fa47.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.27+commit.91d4fa47.js`| `0.3.6-nightly.2016.08.27+commit.91d4fa47.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.29+commit.b8060c55.js`| `0.3.6-nightly.2016.08.29+commit.b8060c55.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.29+commit.b8060c55.js`| `0.3.6-nightly.2016.08.29+commit.b8060c55.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.30+commit.cf974fd1.js`| `0.3.6-nightly.2016.08.30+commit.cf974fd1.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.30+commit.cf974fd1.js`| `0.3.6-nightly.2016.08.30+commit.cf974fd1.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.31+commit.3ccd1986.js`| `0.3.6-nightly.2016.08.31+commit.3ccd1986.Emscripten.clang`      |
| `soljson-v0.3.6-nightly.2016.8.31+commit.3ccd1986.js`| `0.3.6-nightly.2016.08.31+commit.3ccd1986.Emscripten.clang`      |

I have also detected another variant that triggers a crash:
| File name                                            | Reported version                                                 |
|------------------------------------------------------|------------------------------------------------------------------|
| `soljson-v0.3.4-nightly.2016.6.6+commit.e97ac4fb.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.6+commit.e97ac4fb.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.093790d7.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.093790d7.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.ccddd6fd.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.ccddd6fd.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.d593166d.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |
| `soljson-v0.3.4-nightly.2016.6.8+commit.d593166d.js` | `0.3.4-0/Release-Emscripten/clang/Interpreter`                   |

axic: The `0.3.4-0` case is partially handled, for `0.1.3-0` and `0.3.5-0`. For both it was understood that the commit hash was set to 0.

However based on your comment it seems that happened for nightlies only? Can you check 0.1.3 and 0.3.4 what they output?

axic:
@cameel updated, how about now?

translate.js
axic: @cameel you may have missed these follow up questions

cameel: Oh sorry, I was sure I answered everything here. Apparently missed your question about releases.

Releases obviously do not have this problem because they do not include a date. You can check that on the list I prepared in the past when I was rebuilding binaries: https://github.com/ethereum/solc-bin/pull/64#issuecomment-711221554

translate.js
```diff
@@ -18,9 +23,17 @@ function versionToSemver (version) {
...
     return '0.1.3';
   }
+  if (version.indexOf('0.3.4-0') !== -1) {
+    return '0.3.4-nightly';
+  }
   if (version.indexOf('0.3.5-0') !== -1) {
     return '0.3.5';
   }
+  // This parses the obsolete nightly style where the date can have leading zeroes.
+  const nightlyParsed = version.match(/^([0-9]+\.[0-9]+\.[0-9]+)-nightly\.([0-9]+)\.0?([1-9])\.0?([1-9])(.*)$/);
```
cameel, line 33: The `([0-9]+)\.0?([1-9])\.0?([1-9])(.*)` does not look quite right to me. Or am I missing something?

It looks to me like it would not accept two-digit month and day values that don't start with `0`. Like `12` or `27`. The test does seem to parse `2016.08.27` correctly though. Why?
- Is it because it pulls the final `7` into the `.*` part?
- Will it accept the `12` in `0.4.8-nightly.2016.12.16+commit.af8bc1c9.Emscripten.clang`?

r0qs: I think you can fix it by doing this instead:
```javascript
const nightlyParsed = version.match(/^([0-9]+\.[0-9]+\.[0-9]+)-nightly\.([0-9]+)\.0?([1-9]|1[012])\.0?(3[01]+|[12]?[0-9])(.*)$/);
```
So the one and two-digits days and months (excluding leading zeroes) are included.

cameel: @r0qs Do you want to take over reviewing this?

r0qs: @cameel I took a look in this PR, and my suggestion actually does not work for cases like: `2018.2.32`, which would wrongly match: 2018, 2, and 3 instead of 32. Although 32 is not a valid day, the idea here is not to validate the date but only to match broken versions that contain the leading zeros. If a match does not happen, it will just return the version, assuming it is semver compatible.

So I changed @axic regex slightly to support cases like: `2016.12.06`

matheusaaguiar (approved):
It looks to me that the regex now correctly match and exclude the leading zeroes.
Needs a rebase though.

axic:
@cameel maybe this is still useful :)

PR #565, opened 2021-11-18
Fix smtchecker callback tests and add Eldarica

leonardoalt:
Besides fixing things, the nicest thing this PR enables is to use alternative solvers much more easily in a different script, like this:
https://gist.github.com/leonardoalt/70d659eb0ec2f0e82b1fa5681d5ce3e8


smtchecker.js
```diff
@@ -25,10 +25,10 @@ function handleSMTQueries (inputJSON, outputJSON, solver) {
   return inputJSON;
 }
 
-function smtCallback (solver) {
+function smtCallback (solverFunction, solver) {
   return function (query) {
     try {
-      var result = solver(query);
+      var result = solverFunction(query, solver);
```
leonardoalt, line 31: This makes it possible to choose the solver from the callers side.

smtsolver.js
```diff
@@ -1,38 +1,41 @@
...
   {
     name: 'z3',
+    command: 'z3',
     params: '-smt2 rlimit=20000000 rewriter.pull_cheap_ite=true fp.spacer.q3.use_qgen=true fp.spacer.mbqi=false fp.spacer.ground_pobs=false'
   },
+  {
+    name: 'Eldarica',
+    command: 'eld',
+    params: '-horn -t:' + (timeout / 1000)
+  },
```
leonardoalt, line 19: Just adding Eldarica for easy future use, it's currently not used for the tests.

test/smtcallback.js
```diff
@@ -132,8 +132,10 @@ tape('SMTCheckerCallback', function (t) {
       return;
     }
 
-    if (smtsolver.availableSolvers === 0) {
-      st.skip('No SMT solver available.');
+    // For these tests we actually need z3/Spacer.
+    const z3HornSolvers = smtsolver.availableSolvers.filter(solver => solver.command === 'z3');
+    if (z3HornSolvers.length === 0) {
```
leonardoalt, line 137: Not to change things too much, we require z3 to run these tests.

test/smtcallback.js
```diff
@@ -167,9 +169,8 @@ tape('SMTCheckerCallback', function (t) {
       if (source.includes(option)) {
         let idx = source.indexOf(option);
         if (source.indexOf(option, idx + 1) !== -1) {
-          st.skip('SMTEngine option given multiple times.');
-          st.end();
-          return;
```
leonardoalt, line 172: This was actually making none of the actual smtcheckertests to run lol

test/smtcallback.js
```diff
@@ -211,15 +212,18 @@ tape('SMTCheckerCallback', function (t) {
       // `pragma experimental SMTChecker;` was deprecated in 0.8.4
       if (semver.gt(solc.semver(), '0.8.3')) {
         let engine = test.engine !== undefined ? test.engine : 'all';
-        settings = { modelChecker: { engine: engine } };
+        settings = { modelChecker: {
+          engine: engine,
+          solvers: [ 'smtlib2' ]
```
leonardoalt, line 217: This was also wrong before, we need to force the callback to be used.

test/smtcallback.js
```diff
@@ -211,15 +212,18 @@ tape('SMTCheckerCallback', function (t) {
...
+        } };
       }
       var output = JSON.parse(solc.compile(
         JSON.stringify({
           language: 'Solidity',
           sources: test.solidity,
           settings: settings
         }),
-        { smtSolver: smtchecker.smtCallback(smtsolver.smtSolver) }
+        { smtSolver: smtchecker.smtCallback(smtsolver.smtSolver, z3HornSolvers[0]) }
```
leonardoalt, line 226: And here we use z3 if available.

smtsolver.js
```diff
@@ -56,5 +66,5 @@ function solve (query) {
 
 module.exports = {
   smtSolver: solve,
-  availableSolvers: solvers.length
+  availableSolvers: solvers
```
axic, line 69: This is still kind of a breaking change, given we export `smtsolver`.

smtsolver.js
```diff
@@ -44,7 +52,9 @@ function solve (query) {
     if (
       !solverOutput.startsWith('sat') &&
       !solverOutput.startsWith('unsat') &&
-      !solverOutput.startsWith('unknown')
+      !solverOutput.startsWith('unknown') &&
+      !solverOutput.startsWith('(error') &&
```
axic, line 56: What returns `(error`?

test/smtcallback.js
```diff
@@ -211,15 +212,22 @@ tape('SMTCheckerCallback', function (t) {
       // `pragma experimental SMTChecker;` was deprecated in 0.8.4
       if (semver.gt(solc.semver(), '0.8.3')) {
         const engine = test.engine !== undefined ? test.engine : 'all';
-        settings = { modelChecker: { engine: engine } };
+        settings = {
+          modelChecker: {
+            engine: engine,
+            solvers: [
+              'smtlib2'
```
axic, line 219: Why is this test forcing smtlib2?

test/smtcallback.js
axic: This is the main fix, I reckon?

smtsolver.js
leonardoalt: Eldarica returns error strings in that format

smtsolver.js
leonardoalt: and that's ok..? :D

test/smtcallback.js
leonardoalt: to not use the z3 embedded in soljson, and rather use a local solver via the callback

test/smtcallback.js
leonardoalt: The `solverFunction` and `solver` fixes in `smtsolver.js` too, solvers were actually not run

smtsolver.js
axic: @cameel ?

smtsolver.js
axic: Bruh.

leonardoalt: To make them look like smtlib sexpr I guess

smtsolver.js
cameel: This does seem breaking. Unless the whole feature is currently broken enough that no one's really using it. Are people using it in practice?

Maybe we could just add a new member instead?

smtsolver.js
cameel: It's just the few starting characters so I suppose the `error` is followed by more info and is wrapped in a parenthesis together with it?

leonardoalt: Yep, like `(error could not generate model)` or something

smtsolver.js
leonardoalt: > Unless the whole feature is currently broken enough that no one's really using it

it is completely broken, because the solver passing is broken

cameel: In that case I think we're fine changing this. It's not like we'll make anyone's code any more broken than it is already :P

smtsolver.js
cameel: Might be a good idea to add an example in a comment next to it to make it less of a WTF for anyone who spots this later, but other than that, looks like the code is fine here.

smtsolver.js
axic: Oh yeah I remember we discussed to document this in the README and/or Solididty Changelog.

test/smtcallback.js
leonardoalt: done

smtsolver.js
leonardoalt: Sure

leonardoalt:
@cameel @axic I added the requested comments, plus added a piece of code I had locally that increases the output buffer of the child solver process. This is needed when a solver outputs a huge counterexample or proof.

smtsolver.js
```diff
@@ -44,7 +55,9 @@ function solve (query) {
...
+      !solverOutput.startsWith('error') // Eldarica reports errors in an sexpr, for example: '(error "Failed to reconstruct array model")'
```
axic, line 60: You added this to the worng line 😅

leonardoalt: lol true

smtsolver.js
axic: Both actually.

leonardoalt: added to solc-js README

README.md
```diff
@@ -148,32 +148,47 @@ The ``smtSolver`` callback function is used to solve SMT queries generated by
...
+configurations, and relies on Z3, Eldarica or CVC4 being installed locally.  It
+exports the list of locally found solvers and a function that invokes a given
+solver.
 
 #### Example usage with smtSolver callback
 
 ```javascript
 var solc = require('solc');
-var smt = require('smtsolver');
+const smtchecker = require('smtchecker.js');
```
axic, line 163:
```suggestion
const smtchecker = require('solc/smtchecker');
```

README.md
```diff
@@ -148,32 +148,47 @@ The ``smtSolver`` callback function is used to solve SMT queries generated by
...
+const smtsolver = require('smtsolver');
```
axic, line 164:
```suggestion
const smtsolver = require('solc/smtsolver');
```

README.md
```diff
@@ -148,32 +148,47 @@ The ``smtSolver`` callback function is used to solve SMT queries generated by
 Solidity's SMTChecker.  If you have an SMT solver installed locally, it can
```
axic, line 148: Should state what solc-js version this started to work and leave a giant *experimental* disclaimer here that the API can change.

axic:
I'm leaning towards going in with the breaking changes, because the SMT Checker is still experimental. We can also change this code after merging to be backwards compatible before the next solc-js release.

cc @chriseth @cameel

axic (approved):
Code-wise this looks good to me, but we may need to revisit keeping "backwards compatibility" before the next release (0.8.12).

README.md
```diff
@@ -144,36 +144,54 @@ for (var contractName in output.contracts['test.sol']) {
...
+Since version 0.5.1, the ``smtSolver`` callback function is used to solve SMT queries generated by
 Solidity's SMTChecker.  If you have an SMT solver installed locally, it can
 be used to solve the given queries, where the callback must synchronously
 return either an error or the result from the solver.  A default
-``smtSolver`` callback is distributed by ``solc-js``, which relies on either
-Z3 or CVC4 being installed locally.
+``smtSolver`` callback is included in this package via the module
+``smtchecker.js`` which exports the ``smtCallback`` function that takes 1) a
+function that takes queries and returns the solving result, and 2) a solver
+configuration object. The module ``smtsolver.js`` has a few pre defined solver
```
cameel, line 154:
```suggestion
configuration object. The module ``smtsolver.js`` has a few predefined solver
```

README.md
```diff
@@ -144,36 +144,54 @@ for (var contractName in output.contracts['test.sol']) {
...
+      solvers: [ "smtlib2" ]
     }
   }
 };
 
 var output = JSON.parse(
-  solc.compile(JSON.stringify(input), { smtSolver: smt.smtSolver })
+  solc.compile(
+    JSON.stringify(input),
+    { smtSolver: smtchecker.smtCallback(smtsolver.smtSolver, smtsolver.availableSolvers[0]) }
```
cameel, line 188: Maybe we should show an example of how to pass in a custom function as a callback?

Otherwise, if user is expected to use `smtCallback()`, why don't we just expect its parameters instead of a full callback?

smtsolver.js
```diff
@@ -3,37 +3,48 @@ const execSync = require('child_process').execSync;
...
   let solverOutput;
   try {
     solverOutput = execSync(
-      solvers[0].name + ' ' + solvers[0].params + ' ' + tmpFile.name, {
-        stdio: 'pipe'
+      solver.command + ' ' + solver.params + ' ' + tmpFile.name, {
+        encoding: 'utf8',
+        maxBuffer: 1024 * 1024 * 1024,
+        stdio: 'pipe',
+        timeout: timeout // Enforce timeout on the process, since solvers can sometimes go arount it.
```
cameel, line 47:
```suggestion
        timeout: timeout // Enforce timeout on the process, since solvers can sometimes go around it.
```

test/smtcallback.js
cameel: I think we should make it opt-out instead of opt-in. With some equivalent of the `--no-smt` flag. Otherwise it's too easy for such tests to never run in CI.

test/smtchecker.js
```diff
@@ -1,25 +1,107 @@
...
+const semver = require('semver');
+const solc = require('../index.js');
 const smtchecker = require('../smtchecker.js');
+const smtsolver = require('../smtsolver.js');
 
+const preamble = 'pragma solidity >=0.0;\n// SPDX-License-Identifier: GPL-3.0\n';
+//
 tape('SMTChecker', function (t) {
+  // We use null for `solverFunction` and `solver` when calling `handleSMTQueries`
+  // because theses tests do not call a solver.
```
cameel, line 11:
```suggestion
  // because these tests do not call a solver.
```

leonardoalt:
@cameel 

> I think we should make it opt-out instead of opt-in. With some equivalent of the --no-smt flag. Otherwise it's too easy for such tests to never run in CI.

The annoying part about this is that the solc-js CI tests run a bunch of old versions that have no idea about this, I think it'd look quite ugly to just disable it in all of them

test/smtchecker.js
```diff
@@ -1,25 +1,107 @@
...
+  t.test('smtCallback should return type function', (st) => {
+    const response = smtchecker.smtCallback(() => {});
+    st.equal(typeof response, 'function');
+    st.end();
+  });
+
+  t.test('smtCallback should error when passed parser fails', (st) => {
+    const callback = smtchecker.smtCallback((content) => { throw new Error(content); });
+    // eslint-disable-next-line node/no-callback-literal
+    const response = callback('expected-error-message');
```
cameel, line 43: Is this complaining that the variable is called `callback` and that's reserved on node? Why not just call it something else and avoid the warning? `testCallback` or something.

leonardoalt: I copied this from https://github.com/ethereum/solc-js/pull/566

leonardoalt: Will change

test/smtchecker.js
```diff
@@ -1,25 +1,107 @@
...
+
+    const newInput = smtchecker.handleSMTQueries(input, output, smtsolver.smtSolver, z3[0]);
+    st.notEqual(newInput, null);
+
+    const newOutput = JSON.parse(solc.compile(JSON.stringify(newInput)));
+    st.ok(newOutput);
+
+    const errors = newOutput.errors;
+    st.equal(errors.length, 1);
+    st.equal(errors[0].errorCode, '6328');
```
cameel, line 103: Has this been tested with a pre-release binary of the compiler? I'm mostly worried about the pre-release warning that is always issued. It could add an extra error here and solc-js test in Solidity repo would then fail once we merge this.

Overall, I think that it would be a good idea to create a temporary branch in the main repo, switch the solc-js test to this PR and see if it passes. In the past we used to have breakage sometimes after merging stuff in solc-js because of stuff like this.

cameel:
> The annoying part about this is that the solc-js CI tests run a bunch of old versions that have no idea about this, I think it'd look quite ugly to just disable it in all of them

Maybe we could just have it wrapped in some helper that gets a version and tells us if it supports a modern version of the SMT stuff?

In any case, that's not a huge blocker for me, could be done separately. Would be nice to have but this PR has been waiting long enough :)

cameel (approved):
I found mostly typos and added a some suggestions about overall I see nothing wrong here.

As I said in one of my comments, I'd recommend testing it against solc-js tests in the main repo before we merge to avoid breakage.

test/smtchecker.js
leonardoalt: Right, I'll switch this to just search for that error instead, safer.

README.md
leonardoalt: It's just way more convenient, there's 3 layers of different stuff here

test/smtchecker.js
leonardoalt: Yep, just changed it to filter the errors by the error code, and ran it with a pre-release binary that I compiled locally.

leonardoalt:
Updated the small things.
I would prefer to leave the `--no-smt` flag to a different PR to get this one merged. The tests are working with a pre-release binary.

README.md
cameel: Then maybe we should show both. I mean, this is the only reference doc we have for the library. Having an easy to run example is good but we should also say something about what the callback looks like. Looking quickly at README, I don't think it even mentions what args the actual callback takes. Only the args of this helper.

leonardoalt: Ok, makes sense. Next PR tho? :D

cameel: Fine. Let's get this one merged :)

cameel:
LGTM

Issue #174, opened 2018-01-05
Properly solve the greedy exception catcher problem


axic:
Apparently there is a way in Emscripten to disable the "greedy exception catcher": https://github.com/kripken/emscripten/blob/f585dcbc2d929ef8b8bc6813e0710ec3215ac0b1/src/settings.js#L325-L328 (Officially suggested here: https://github.com/kripken/emscripten/issues/5957)

Found by @hugo-dc in https://github.com/ethereumjs/rustbn.js/pull/11

Properly fixes #34, #44 and #113.

axic:
Once compiled with that we could remove this code: https://github.com/ethereum/solc-js/blob/master/solcjs#L9-L11

Issue #34, opened 2016-08-01
Improve greedy exception catcher of Emscripten


axic:
Follow up from https://github.com/chriseth/browser-solidity/issues/167.

The Emscripten preamble inserts an exception catcher for _uncaughtException_ in Node.js.  This rethrows all exceptions, bar the ExitCode one.

The rethrow causes stack traces to be confusing long.

PR #569, opened 2021-12-03
Remove the workaround for greedy exception catcher

cameel:
This PR removes the old workaround that was added in #34 to solve https://github.com/chriseth/browser-solidity/issues/167.

The issue was solved properly by adding `NODEJS_CATCH_EXIT=0` flag to our emscripten build in 0.4.20 (https://github.com/ethereum/solidity/pull/3477). According to https://github.com/ethereum/solc-js/issues/174#issuecomment-358016947 the workaround is no longer needed.

### DIFF
--- a/solcjs
+++ b/solcjs
@@ -1,17 +1,11 @@
 #!/usr/bin/env node
 
-// hold on to any exception handlers that existed prior to this script running, we'll be adding them back at the end
-var originalUncaughtExceptionListeners = process.listeners("uncaughtException");
-
 var fs = require('fs-extra');
 var os = require('os');
 var path = require('path');
 var solc = require('./index.js');
 var smtchecker = require('./smtchecker.js');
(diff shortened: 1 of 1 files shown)


cameel:
@axic https://github.com/ethereum/solc-js/issues/174#issuecomment-358016947 says we can safely remove it but won't that break on pre-0.4.20 binaries?

Well, wasm rebuilds were done long after 0.4.20 so I think they do not have this problem and also the change only affects the `solcjs` tool, not whole library so I created the PR anyway.

axic:
Actually we probably can't remove it if we want to keep supporting old compiler versions with this CLI tool. Those binaries still need it 😢

cameel:
Oh well, I was afraid that this would be the end result. Yeah, I think we want to support them, I guess there's no magic way to do it without this workaround :)

I'm closing it then. We can always revive if we decide to merge it after all.

PR #572, opened 2021-12-14
Remove fs-extra.

chriseth:


chriseth:
It seems this is not working on nodejs v8

MicahZoltu:
Given that NodeJS 8 is past end of life (no longer receiving maintenance updates), my vote is to just drop v8 support.

https://nodejs.org/en/about/releases/

Ruby6789:
#572

.circleci/config.yml
```diff
@@ -55,10 +54,6 @@ jobs:
     <<: *node-base
     docker:
       - image: circleci/node:6
-  node-v8:
```
cameel, line 58: If you're removing this I think there's no point in keeping `node-v4` and `node-v6` above.

cameel (approved):
Looks good overall, just v4 and v6 should be removed from CI too.

Also, we should probably mention it in the changelog.

chriseth:
Which changelog? The main solidity changelog?

cameel (approved):

cameel:
Yes, isn't that changelog meant to cover both solidity and solc-js? Last time I did a bigger change here, you said to add it to the changelog in the main repo.

chriseth:
Changelog: https://github.com/ethereum/solidity/pull/12438

axic:
We actually discussed last week with @cameel that dropping nodev8 support may be "breaking" in terms of solc-js, hence we did not proceed yet with the eslint changes. I think this must be highlighted well in the changeling, and ideally would have only been done in 0.9.0.

chriseth:
When we drop ubuntu support for releases, we also don't do that in breaking changes. nodejs v8 has been dead for months and anyone remotely concerned about security should have switched a long time ago.

cameel:
Yeah, I agree that it would normally be best done in a breaking release but I think in this case it's so old that probably no one cares. It will also let us proceed with the Typescript migration before people willing to contribute now end up losing interest :)

PR #573, opened 2021-12-19
Removes dependency on fs-extra.

MicahZoltu:
This should remove ~16 of the 25 dependencies of this project, which reduces the attack surface area and gets us down to only direct dependencies I believe.

See https://github.com/ethereum/solc-js/issues/361#issuecomment-993230407 for additional discussion.

I went ahead and removed the path check per thought from @chriseth, but I'm happy to put it back in if desired.


MicahZoltu:
Bleh, just noticed #573.  Closing as duplicate.

PR #586, opened 2022-01-24
fix: os.platform is a function call not a property

stephensli:


stephensli:
could we have some tolerance on the coverage? -0.005% is a little extreme.

axic (approved):

PR #566, opened 2021-11-18
Typescript Base Support

stephensli:
# Base Typescript Support  #287 

This pull request includes the base requirements and minor changes to keep the current application code, test code, and process while supporting Typescript, Typescript compiling, and future support for adding types.

### Files

All files have been moved from `.js` to `.ts`. This includes providing an extension type for the binary file `solcjs` which is now `solc.ts` which will later be transpiled into `solc.js`. The required package.json references have been updated where applicable to ensure packaging and binary referencing when linking work as expected.

### Linting

Linting has been directly upgraded from just `standard/standard` to use `eslint` which is directly supported within Typescript, allowing code style enforcement of `standard/standard` while also locating and flagging possible code errors. Linting has been run throughout to change `var` to `cont` where applicable. 

### Post Build Script

A new script file exists within the `build` folder for post-build. This correctly sets the required permissions on the `solc.js` command-line binary reference. Allowing testing to function correctly and direct execution to work as expected.

### Testing

Testing is done against the transpiled code after running through the Typescript compiler. This is due to the usage of dynamic importing being performed to pull in different binary versions. Each time this is performed the typescript compiler will attempt to parse and validate the entire file every single time. This increases testing time 10 fold and should be avoided. Testing against the built code allows testing to work as it does now.

When running the test command, additional resource files are copied into the `dist` folder to allow testing to function correctly. This is due to Typescript not including those files in the build process. 

### Packaging

When publishing or packaging, the dist files are used instead of the source typescript files as expected. You can test the content of the deployment or release by running the following `npm pack`.

### What's next?

Once merged, this allows the introduction of typing to reduce errors and improve code quality. This can go in hand with enabling typescript declaration and configuration of the package.json to use said built types. Allowing users of the application to gain access to these types after installing the package.

PR #587, opened 2022-01-24
Rename solcjs to solc.js internally and apply lint changes

axic:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Solidity compiler",
   "main": "index.js",
   "bin": {
-    "solcjs": "solcjs"
+    "solcjs": "solc.js"
```
axic, line 7: I wonder if the `files` list below also needs an update?

axic:
@stephensli pulled out this monster change from #566.

package.json
stephensli: Yes I believe so, 

This would be the file which will be in the release, you can run `npm pack` to simulate a release package and see the content. Make sure everything you are expecting to also be in that package.

https://docs.npmjs.com/cli/v7/commands/npm-pack

axic: It seems both versions work as maybe it looks up the dependency. At least I got the same tarball `files: [ .. "solcjs" .. ]` and `files: [ .. "solc.js" .. ]`.

stephensli: It would make sense that it pulls the bin in as well, probably worth being clear with the definition in the files section as well for clarity.

axic: Would you suggest `solc.js` then in the file list?

stephensli: yeah since solc.js is the expected file we want.

solc.js
```diff
@@ -171,14 +166,15 @@ if (!options.basePath && options.includePath && options.includePath.length > 0)
...
-    if (!includePath)
-      abort('Empty values are not allowed in --include-path.');
+if (options.includePath) {
+  for (const includePath of options.includePath) {
+    if (!includePath) { abort('Empty values are not allowed in --include-path.'); }
+  }
+}
 
-var sources = {};
+const sources = {};
```
leonardoalt, line 175: let?

axic: I was also confused but apparently the initial type assignment is const and not the values of it 🤷

solc.js
```diff
@@ -193,37 +189,36 @@ const cliInput = {
   settings: {
     optimizer: {
       enabled: options.optimize,
-      runs: options.optimizeRuns,
+      runs: options.optimizeRuns
```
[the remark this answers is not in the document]
axic: eslint doesnt 😬

solc.js
leonardoalt: wat ok

axic: See also https://github.com/ethereum/solc-js/pull/566#discussion_r790301655

leonardoalt (approved):

solc.js
stephensli: Yeap 

```js
// valid
const person = { name: "stephen" }
person.name = "leon"

// invalid
const personTwo = { name: "stephen" }
personTwo = ""

// valid 
let personThree = { }
personThree = ""

File: solc.js (hunk 3 above)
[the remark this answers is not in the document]
Comment (stephensli) [reply solc.js:192]: If you want dangling commas the rules can be adjusted to enforce or allow them.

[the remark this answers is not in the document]
Comment (axic) [reply solc.js:192]: I think it is fine, let's focus on getting the changes in and revisit linting after most of the typescript work is done.

File: solc.js (hunk 2 above)
Comment (cameel) [reply solc.js:175]: > ``` javascript
> // valid
> const person = { name: "stephen" }
> person.name = "leon"
>
> // invalid
> const personTwo = { name: "stephen" }
> personTwo = ""
> ```

Yeah, I kinda expected that.

> ``` javascript
> // valid 
> let personThree = { }
> personThree = ""
> ```

wat

JS was supposed to be only the good parts now :P

File: solc.js (hunk 3 above)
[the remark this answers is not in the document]
Comment (cameel) [reply solc.js:192]: Yeah, we can revisit later. But I do want dangling commas too :P

File: solc.js [hunk 4]
```diff
@@ -104,58 +105,52 @@ function makeSourcePathRelativeIfPossible(sourcePath) {
...
-function reformatJsonIfRequested(inputJson) {
+function reformatJsonIfRequested (inputJson) {
   return (program.prettyJson ? toFormattedJson(JSON.parse(inputJson)) : inputJson);
 }
 
-var callbacks = undefined
-if (options.basePath || !options.standardJson)
-  callbacks = {'import': readFileCallback};
+let callbacks;
+if (options.basePath || !options.standardJson) { callbacks = { import: readFileCallback }; }
```
cameel, line 124: Oh, does it really reformat like this automatically?
Can't say I like this. Makes the code less readable in many cases.

test/compiler.js
```diff
@@ -289,7 +289,7 @@ function runTests (solc, versionText) {
           'cont.sol': 'import "lib.sol"; contract x { function g() public { L.f(); } }'
         };
         st.throws(function () {
-          solc.lowlevel.compileCallback(JSON.stringify({ sources: input }), 0, "this isn't a callback");
+          solc.lowlevel.compileCallback(JSON.stringify({ sources: input }), 0, 'this isn\'t a callback');
```
cameel, line 292: This one also feels like a downgrade.

cameel (approved):
Well, I can't say I like all the formatting choices but I guess this is automatic and does not break anything so I'm approving anyway.

If it's configurable though I'd really like to change the two bits I pointed out.

test/compiler.js
axic: We do this in other places too.

solc.js
axic: If we change the style I'd suggest in a different PR because that will apply to every single file. In this PR the only change really is enabling lint on `solcjs`.

axic:
I'm merging this for the moment, and we can iterate on style separately.

solc.js
stephensli: :D 

There is barely any reason to use let over const only in situations like

```js
let count = 0;
count += 1;
```

PR #591, opened 2022-01-25
Fix: Ensure that tests work even if /tmp is not accessible

### LINKED ISSUE #588: Ensure that tests work even if `/tmp` is not accessible
`test/compiler.js` assumes that `/tmp` is always accessible. We should use a function that creates a temporary dir instead.

_Originally posted by @cameel in https://github.com/ethereum/solc-js/pull/566#discussion_r762097623_

adi611:
Fixes #588.

- [x] Created a temporary directory using `tmp` package inside function `createTempDir()`
- [x] Returned the name of temporary directory from the function `createTempDir()`
- [x] Replaced `/tmp` with name of created temporary directory `tempDir`

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   },
   "devDependencies": {
     "@types/node": "^16.11.7",
+    "@types/tmp": "^0.2.3",
     "@types/semver": "^7.3.9",
     "@types/tape": "^4.13.2",
     "@typescript-eslint/eslint-plugin": "^5.8.0",
--- a/test/compiler.ts
+++ b/test/compiler.ts
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import tape from 'tape';
 import * as semver from 'semver';
+import * as tmp from 'tmp';
 import solc from '../';
 import linker from '../linker';
 import { execSync } from 'child_process';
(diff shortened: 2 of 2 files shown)


test/compiler.ts
```diff
@@ -8,6 +8,12 @@ import wrapper from '../wrapper';
 
 const noRemoteVersions = (process.argv.indexOf('--no-remote-versions') >= 0);
 
+// use a function that creates a temporary dir instead of /tmp
+function createTempDir () {
+  const tmpDir = require('tmp').dirSync({ unsafeCleanup: true });
```
axic, line 13: Please use an import statement above and not `require` here.

adi611: Ok

adi611:
It looks like there were some changes to `smtsolver.ts`

package.json
```diff
@@ -59,6 +59,7 @@
   },
   "devDependencies": {
     "@types/node": "^16.11.7",
+    "@types/tmp": "^0.2.3",
```
axic, line 62: Was `tmp` not used in any other place we just had it `dependencies`?

axic:
Please squash commits to a single one and use rebase instead of merge commit.

package.json
adi611: Sorry but can you please be more clear about it?

axic: We have `tmp` in the `dependencies` in `package.json`. I suppose the typing package is needed if it is actually used within Typescript. I wonder why this one was not needed so far, which could mean that we do not actually use the package prior to this change.

adi611: This is actually quite funny lol.

adi611:
Hi I was wondering whether I need to do any more changes for this to get merged.

cameel (changes requested):
I'd adjust this a bit (see my comment) but other than that it looks fine to me and I can merge it when you apply these changes.

adi611:
Done @cameel

cameel:
Looks good, but can you also squash commits before I merge this? Our convention is that a every commit should be a single, logical change without going back and forth. Review fixes should be just squashed into the original commits.

adi611:
> Looks good, but can you also squash commits before I merge this? Our convention is that a every commit should be a single, logical change without going back and forth. Review fixes should be just squashed into the original commits.

Done @cameel

cameel (approved):

cameel:
Great. Thanks for the contribution!

adi611:
Thanks for helping me along the way.

PR #593, opened 2022-01-25
fix: export all properties under the root instead of default

stephensli:
By exporting under the root, javascript importing modules will continue
to use it as expected without first having to index `.default.`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,26 @@
 import wrapper from './wrapper';
 
 const soljson = require('./soljson.js');
+const wrapped = wrapper(soljson);
 
-export default wrapper(soljson);
+const {
+  version,
+  semver,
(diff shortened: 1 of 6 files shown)


ekpyron:
Ok, since https://github.com/ethereum/solidity/pull/12585 passes its tests relative to this change, I'm merging this, but @cameel and @axic should still have a look at it as well.

cameel:
Looks like this PR was enough to make it work with Truffle. For me it still fails with Hardhat though. That's because of the `dist/` directory that's now in the package.

I rebased #594 on top of this and all its tests should be passing if things are fixed. The truffle ones just need further adjustments so please ignored them until I fix them but I think the issue in the Hardhat ones is real.

PR #594, opened 2022-01-25
CI tests based on Truffle and Hardhat

cameel:
~Depends on #596. Should not be merged before that PR.~ Merged.

This PR adds CI jobs that run latest Truffle and Hardhat with injected solc-js. This runs their test suites and also tries to compile a sample project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,12 @@ workflows:
 version: 2.1
 
 commands:
+  show-npm-version:
+    steps:
+      - run:
+          name: Versions
+          command: npm version
+
   install-dependencies:
     parameters:
       cache-id:
@@ -48,9 +54,7 @@ jobs:
         type: boolean
         default: false
     steps:
-      - run:
-          name: Versions
-          command: npm version
+      - show-npm-version
       - checkout
       - install-dependencies:
           cache-id: solc-js
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,34 @@ workflows:
       - node-v17
 
 version: 2.1
+
+commands:
+  install-dependencies:
+    parameters:
+      cache-id:
+        type: string
+      path:
+        type: string
+        default: .
+      package-manager:
+        type: string
+        default: npm
+      dependency-file:
+        type: string
+        default: package.json
+    steps:
+      - restore_cache:
+          key: <<parameters.cache-id>>-dependency-cache-v2-{{ .Environment.CIRCLE_JOB }}-{{ checksum "<<parameters.path>>/<<parameters.dependency-file>>" }}
+      - run:
+          name: "<<parameters.package-manager>> install in <<parameters.path>>"
+          command: |
+            cd "<<parameters.path>>"
+            [[ -e node_modules/ ]] || <<parameters.package-manager>> install
+      - save_cache:
+          key: <<parameters.cache-id>>-dependency-cache-v2-{{ .Environment.CIRCLE_JOB }}-{{ checksum "<<parameters.path>>/<<parameters.dependency-file>>" }}
+          paths:
+            - "<<parameters.path>>/node_modules/"
+
 jobs:
   node-base: &node-base
     working_directory: ~/solc-js
@@ -24,8 +52,8 @@ jobs:
           name: Versions
           command: npm version
       - checkout
-      - restore_cache:
-          key: dependency-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "package.json" }}
+      - install-dependencies:
+          cache-id: solc-js
       - run:
           name: install-npm
           command: npm install
@@ -41,10 +69,6 @@ jobs:
             - run:
                 name: coveralls
                 command: npm run coveralls
-      - save_cache:
-          key: dependency-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "package.json" }}
-          paths:
-            - ./node_modules
 
   node-v10:
     <<: *node-base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,10 @@ workflows:
       - node-v16:
           run_coveralls: true
       - node-v17
+      - hardhat-core-default-solc
+      - hardhat-core-latest-solc
+      - hardhat-sample-project
+      - truffle-sample-project
 
 version: 2.1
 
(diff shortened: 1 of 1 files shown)


ekpyron:
Changes to solc-js are relatively rare, so there's not that much need to run this stuff quickly... don't you think we could just make it easy to configure the solc-js branch in the solidity repo and just check that out and run ``test/externalTests.sh`` on that?
Otherwise we double the maintenance cost of the external tests.

cameel:
@ekpyron They are also very frustrating to test. I never feel that my code here is tested well enough not to cause some kind of breakage.

The CI in the main repo is already relatively easy to reconfigure for this (could be easier but still, it's not hard) but it's still a pain. And the major downside to that is that it requires our intervention and is not usable to contributors - and a lot of PRs in solc-js are coming from contributors. The typescript PR is a great example. These issues would have been dealt with long ago but we only discovered them now.

As for maintenance, well, it's a trade-off. The tests here are not really exactly the same thing as in the main repo. They're simplified and, after seeing how they work, I'm considering dropping some of them. Details in a moment.

ekpyron:
Maybe I wasn't clear - I meant we can just run a version of the solidity external tests relative to a solc-js branch in solc-js PRs...
If something breaks due to the solidity side in that, then it needs our intervention anyways...

ekpyron:
But yeah, if we can figure out adding CI runs here in a nice, maybe it's also ok, let's see :-).

cameel:
> I meant we can just run a version of the solidity external tests relative to a solc-js branch in solc-js PRs

Ah, running external tests from Solidity in this repo? That would work but it's not the part that adds that much complexity here. It's that I build Truffle and Hardhat manually to inject latest solc-js into their deps. In the main repo we're not doing that, we're installing them from npm packages.

The Truffle failures we got in https://github.com/ethereum/solidity/pull/12583 were only coming from the fact that even with older solc-js installed from npm, Truffle still expects the emscripten binary to come wrapped in another of copy `solc-js` and that one is the latest. In external tests that's the only point of contact with the latest version. Hardhat takes `soljson.js` binary directly so it did not even fail that way. I think they will both break once latest solc-js hits npm. Tests in the main repo are not enough to show that but the ones I'm adding here do.

So yeah, I don't think that using external tests here would shave off that much code - it would only really replace the final bits in `truffle-sample-project` and `hardhat-sample-project` jobs - but it would still be doable. I considered that but I think these projects are more than we need here. We don't benefit too much from extensive test suites external projects have. Just being able to run compilation and *some* tests is IMO enough and takes much less time in CI. I see it more as a smoke test. Also, they should be much more stable. External projects break all the time. The ones I used here are very simple tutorials that should not break and are trivial to port over breaking changes. They're even all on older solc but the only thing I needed to make them work here was to remove the pragma.

In one case I actually did what you're suggesting. The `solidity-solcjs-ext-test` job is just checking out the Solidity repo and running `externalTests/solc-js/solc-js.sh`. This case does not need Truffle or Hardhat so it does not have the problems I mentioned above.

cameel:
I have removed the `truffle-tests-default-solc` job. It runs Truffle's test suite but it does not pass because of missing stuff. It generally assumes that you have a full setup necessary to use all of Truffle's features, including access to docker, installed vyper binary, etc. And it's unfortunately not cleanly separated so I can't work around it by ignoring some packages. I can easily skip tests for `@truffle/compile-vyper` but the failing ones are also in `@truffle/compile-solidity` and that's the one we'd want the most.

Here's what the job looked like in case we want to add it back later:
```yaml
  truffle-tests-default-solc:
    docker:
      - image: circleci/node:12
    steps:
      - show_npm_version
      - provision_and_package_solcjs
      - provision_truffle_with_packaged_solcjs
      - run:
          name: Run Truffle test suite
          command: |
            cd truffle/
            npx \
              lerna run test \
                --ignore @truffle/compile-vyper \
                --ignore @truffle/debugger \
                --stream \
                --concurrency=2
```

ekpyron:
Yeah, it actually makes *more* sense not to replace solc-js with ``master`` everywhere, since solc-js guarantees forward-compatibility within major version cycles. So it's indeed true that it's more important to test stuff like this here.

.circleci/config.yml
```diff
@@ -41,10 +148,140 @@ jobs:
...
+      - image: circleci/node:16
+    steps:
+      - show_npm_version
+      - provision_and_package_solcjs
+      - provision_hardhat_with_packaged_solcjs
+      - run:
+          name: Run hardhat-core test suite with its default solc
+          command: |
+            cd hardhat/packages/hardhat-core
+            yarn test
```
cameel, line 163: This test is **much** slower than `hardhat-core-latest-solc`. It runs with 0.5.x and runs about 4k tests, which takes ~15 min. `hardhat-core-latest-solc` has 10 times fewer tests and finishes in ~1 min.

Not sure if we should keep it or not. I think it may still be adding some value because I think that tests on 0.8.x do not actually go through all the features. In particular Hardhat's downloader module tests get skipped and that's one of the few things in Hardhat that uses solc-js.

I think that Hardhat prepared `HARDHAT_TESTS_SOLC_PATH` specifically with testing the latest compiler in mind and disabled the extra tests on purpose. Probably because they're not relevant if you're testing solc, but I think some of them do matter for solc-js.

.circleci/config.yml
```diff
@@ -41,10 +148,140 @@ jobs:
...
+  solidity-solcjs-ext-test:
+    docker:
+      - image: circleci/node
+    steps:
+      - show_npm_version
+      - checkout:
+          path: solc-js/
+      - run: git clone --depth 1 "https://github.com/ethereum/solidity" solidity/
+      - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
+      - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)"
```
cameel, line 284: Note that I'm using a nightly binary here to have something that matches the version returned from `scripts/get_version.sh`. Alternatively I could make it work with latest release binary, which is used in other jobs but this was easier and I also think that having something using the nightlies in CI might let us spot more problems.

.circleci/config.yml
ekpyron: Since we don't have a really high volume of changes in solc-js, I guess 15 minutes tests is not *that* bad...

.circleci/config.yml
cameel, line 284: I just realized that this will run with solc-js independently checked out from `master` by the `solc-js.sh` script. I need to adjust `solc-js.sh` in the main repo to accept an existing checkout of solc-js instead. For now I'm removing it from the PR and I'll resubmit it separately.

ekpyron (commented):
From just looking at it I couldn't really say if it properly does what it should - the tests breaking now and hopefully being fixed by an upcoming change in solc-js will give some assurance, though, and in general it looks reasonable to me.

ekpyron (approved):

PR #595, opened 2022-01-26
fix: use esModuleInterop to resolve commonjs default require

stephensli:
By doing so allows the commonjs default imports to work as expected not blocking
backwards compatibility.

reference: https://www.typescriptlang.org/tsconfig#esModuleInterop

This will stop imports like the following from being a issue.

```js
const wrapper = require('solc/wrapper).default;
```

they will now work as expected as

```js
const wrapper = require('solc/wrapper);
```

### DIFF
--- a/translate.ts
+++ b/translate.ts
@@ -193,7 +193,7 @@ function prettyPrintLegacyAssemblyJSON (assembly, source) {
   return formatAssemblyText(assembly, '', source);
 }
 
-export default {
+export = {
   versionToSemver,
   translateJsonCompilerOutput,
   prettyPrintLegacyAssemblyJSON
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,10 @@
     "target": "esnext",
     "module": "commonjs",
     "resolveJsonModule": true,
+    // This is needed for backwards-compatibility, to keep imports of the form `wrapper = require('solc/wrapper)`
+    // working like they did before the TypeScript migration.
+    // TODO: Drop it in the next breaking release.
+    "esModuleInterop": true,
     "outDir": "./dist",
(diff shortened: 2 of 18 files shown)


cameel:
Can you rebase this on #594 to see if tests there will pass?

translate.ts
```diff
@@ -193,8 +193,8 @@ function prettyPrintLegacyAssemblyJSON (assembly, source) {
   return formatAssemblyText(assembly, '', source);
 }
 
-export default {
+export = {
   versionToSemver,
   translateJsonCompilerOutput,
   prettyPrintLegacyAssemblyJSON
-};
+}
```
cameel, line 200: I'd leave these semicolons as is in this PR.

tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "target": "esnext",
     "module": "commonjs",
     "resolveJsonModule": true,
+    "esModuleInterop": true,
```
cameel, line 7: A comment explaining why we're using this option would be nice.

cameel (approved):
I have two minor remarks. Apart from that this looks fine as far as I understand exports in Typescript. And we'll want the code to pass tests from #594 eventually so if anything's wrong, we'll have a chance to correct this.

We need to do something about coverage here. It's weird that it fell by 2% - the PR removes more code than it adds and it's likely code that counts as not covered so I'd expect it to go up instead.

translate.ts
cameel: Fixed.

tsconfig.json
cameel: Done.

cameel (approved):

PR #596, opened 2022-01-26
fix: use npm run publish to release the dist folder directly

stephensli:
~Depends on #595. Should not be merged before that PR.~ Merged.

# Why?

When releasing the package, and users begin to use the application, the following is a example import which can take place.

```js
const wrapper = require('solc/wrapper');
```

With the introduction of the typescript support, the content is built into the dist folder. Resulting in the build folder to look like the following.

```
package.json
dist/wrapper.js
dist/solc.js
```

The following change ensures that we publish the dist folder directly which will result in a flatten published folder like the following 

```
package.json
wrapper.js
solc.js
```

which will allow the code example to function correctly without changing the import path to `solc/dist/wrapper`. This change will go in hand with the following PR #595 to allow the commonjs imports to function as expected without breaking backwards compatibility.

# Release Change

Executing `npm publish` or `npm pack` will now be rejected by the application. This is because we want to pack, build and release the built dist folder. By breaking out the process into `build:tarball` and `publish:tarball` the user has more fine control over the process. With additional support for testing.

Running `npm pack` or `npm publish` will show the following error: Run `npm run build:tarball` or `npm run publish:tarball` to pack or publish the package`

## Commands

`npm run build:tarball` will update the binary, build the project and pack the dist folder. The tarball would be created as solc-{version}
`npm run publish:tarball` will locate the tarball file and publish it. If it is not located then the process is closed and the user is notified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "solc",
   "version": "0.8.11",
   "description": "Solidity compiler",
-  "main": "dist/index.js",
+  "main": "index.js",
   "bin": {
-    "solcjs": "dist/solc.js"
+    "solcjs": "solc.js"
   },
   "scripts": {
     "build": "tsc",
@@ -12,7 +12,10 @@
     "lint": "eslint --ext .js,.ts .",
     "lint:fix": "eslint --fix --ext .js,.ts .",
     "updateBinary": "node build/clean.js && ts-node ./downloadCurrentVersion.ts && ts-node ./verifyVersion.ts",
-    "prepublishOnly": "npm run updateBinary npm && npm run build",
+    "prepack": "node build/pack-publish-block.js",
+    "build:tarball": "npm run updateBinary && npm run build && BYPASS_SAFETY_CHECK=true npm pack ./dist",
+    "publish:tarball": "tarball=$(npm run --silent tarballName) && ls \"$tarball\" && BYPASS_SAFETY_CHECK=true npm publish \"$tarball\"",
+    "tarballName": "jq --raw-output '.name + \"-\" + .version + \".tgz\"' package.json",
     "copyTestFiles": "cp -r ./test/resources ./dist/test/",
     "pretest": "npm run lint && npm run build && npm run copyTestFiles",
     "test": "cd dist && tape ./test/index.js",
@@ -32,15 +35,15 @@
     "node": ">=10.0.0"
   },
   "files": [
-    "dist/abi.js",
-    "dist/index.js",
-    "dist/linker.js",
-    "dist/smtchecker.js",
-    "dist/smtsolver.js",
-    "dist/solc.js",
-    "dist/soljson.js",
-    "dist/translate.js",
-    "dist/wrapper.js"
+    "abi.js",
+    "index.js",
+    "linker.js",
+    "smtchecker.js",
+    "smtsolver.js",
+    "solc.js",
+    "soljson.js",
+    "translate.js",
+    "wrapper.js"
   ],
   "author": "chriseth",
   "license": "MIT",
--- a/abi.ts
+++ b/abi.ts
@@ -58,6 +58,6 @@ function update (compilerVersion, abi) {
   return abi;
 }
 
-export default {
+export = {
   update
 };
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -2,17 +2,19 @@
...
   },
   "scripts": {
     "build": "tsc",
     "postbuild": "node build/postbuild.js",
     "lint": "eslint --ext .js,.ts .",
     "lint:fix": "eslint --fix --ext .js,.ts .",
     "updateBinary": "node build/clean.js && ts-node ./downloadCurrentVersion.ts && ts-node ./verifyVersion.ts",
-    "prepublishOnly": "npm run updateBinary npm && npm run build",
+    "prepublishOnly": "node build/prepublish.js",
+    "prepublish": "npm run updateBinary npm && npm run build && cd dist",
```
cameel, line 16:
```suggestion
    "prepublish": "npm run updateBinary && npm run build",
```
The extra `run` seems like something that should not be there.
Also the final `cd dist` has no effect, right? It can't change directory in your shell.

cameel:
As discussed on Gitter, we need to have a command to pack it into a tarball for testing. This PR should be rebased on top of #594 and it should modify the tests added there to use the pack command.

The pack command should be kept next to the publish command because they must always be updated in sync. I actually think that the `publish` command should just use that same tarball. That way we can be 100% sure we're publishing the same thing we're testing but @stephensli thinks it's not necessary and just overcomplicates things. We'd need a third opinion from @axic here.

Finally, once this gets merged we need to remember to update the [release checklist](https://github.com/ethereum/solidity/blob/develop/ReleaseChecklist.md) to use the new publish command.

build/prepublish.js
```diff
@@ -0,0 +1,6 @@
+const RELEASE = !!(process.env.RELEASE);
+
+if (!RELEASE) {
```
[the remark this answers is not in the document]
stephensli: just for clarity.

```
> !!process.env.test
false
> !process.env.test
true
>
```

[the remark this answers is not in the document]
cameel: I think that `!process.env.RELEASE` is clear enough (it's not like it has a bad name and has to be renamed) and a single negation already gives you a boolean. Adding an intermediate variables actually sometimes makes things less clear because you now have another name to lookup.

But ok, it's just a tiny nitpick so if you really prefer it like that then fine.

build/build-tarball.js
```diff
@@ -0,0 +1,25 @@
...
+    child.stdout.pipe(process.stdout);
+    child.stderr.pipe(process.stderr);
+
+    await new Promise((resolve) => {
+      child.on('close', resolve);
+    });
+  }
+}
+
+main().then().catch(console.error);
```
cameel, line 25: Why do you need all that? What's wrong with putting `npm run updateBinary && npm run updateBinary && npm pack ./dist` in `package.json`? Was there some kind of problem with that?

build/publish-tarball.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const child = spawn('npm', ['publish', tarPath], args);
+
+  child.stdout.pipe(process.stdout);
+  child.stderr.pipe(process.stderr);
+
+  await new Promise((resolve) => {
+    child.on('close', resolve);
+  });
+}
```
cameel, line 28: Here again, it seems to me like you could accomplish the same thing much easier with
```bash
npm publish solc-*.tgz
```
or better this way to make sure we don't try to publish some old tarball lying around:
```bash
npm publish "solc-$(jq --raw-output .version package.json).tgz"
```

stephensli:
@cameel The scripts are built for clarity, maintainability and to remove the chances of mistakes. These are the two single most important part of the application release process. We don't want any mistakes by adding it all directly to package.json.

It also supports additional logging.

cameel:
> The scripts are built for clarity, maintainability and to remove the chances of mistakes.

Sorry, but I really don't see it. It's longer and more complex than the one-liners. To me it seems like a more convoluted way of achieving exactly the same thing.

> It also supports additional logging.

You could do it the way you did previously - with a short stub that either prints a message and exits or lets the command continue. I think that was a better approach.

cameel:
Once this is merged we'll need to merge https://github.com/ethereum/solidity/pull/12595 in the main repo to keep the release checklist up to date.

cameel:
@chriseth I tested the publish process and it worked with a tarball without any problems. Here's the test package I published: [solc-tarball-publish-test__](https://www.npmjs.com/package/solc-tarball-publish-test__).

PR #597, opened 2022-01-27
Test/commonjs package flatten framworks

stephensli:


.circleci/config.yml
```diff
@@ -268,7 +262,7 @@ jobs:
             cd metacoin/
             # `truffle test` compiles the project but artifacts go into /tmp/
             ! [[ -e build/ ]] || false
-            echo "module.exports['compilers'] = {solc: {version: '$(realpath ../solc-js/)'}}" > truffle-config.js
+            echo "module.exports['compilers'] = {solc: {version: '$(realpath ../solc-js/dist)'}}" > truffle-config.js
```
cameel, line 265: I think it would probably be best to change it to:
```suggestion
            echo "module.exports['compilers'] = {solc: {version: '$(realpath node_modules/solc/)'}}" > truffle-config.js
```

This way again we don't have to make an independent decision to use `dist/` but just use whatever gets published.

I think I'll change this in my PR directly.

stephensli: Yeap all these changes should happen in your PR when you introduce the changes.
I just make the change to fix the test and did't know of the correct location it should happen.

cameel: I mean, your change is good and in the correct place. I'm just saying that there's a more general solution and I should have used it.

And yeah, these changes should be in my PR but it makes sense to put them there only after it's on top of your PR, i.e. after yours gets merged.

cameel:
As for your question about the failures:
> I assume the second is failing because Inject local solc-js into Hardhat's dependencies does not inject soljson.js but i'm not sure about the first. Will look at the second tomorrow.

```
1) Compiler downloader
       Downloaded compiler verification
         Shouldn't do anything if the compiler is fine:
     Error: ENOENT: no such file or directory, unlink '/tmp/hardhat-tests-compiler-downloader/wasm/list.json'
  

  2) Compiler downloader
       getDownloadedCompilerPath
         If the compiler already existed
           Should return it if it's passes the verification:
     Error: ENOENT: no such file or directory, stat '/home/circleci/project/hardhat/node_modules/solc/soljson.js'
```

I'm debugging this right now the package gets installed properly. It's just that `soljson.js` disappears during the test. I think Hardhat's test cleanup must be removing it for some reason.

cameel:
I found the problem. Apparently it's because one of the duplicate `solc/` modules that I've been removing to make sure they're not used is actually used in one of the tests (`test/internal/solidity/compiler/downloader.ts`). So basically the safeguard worked as designed, the only real problem was that the effect of this is really weird and disconnected from the cause - WTF does removing `solc/` result in `soljson.js` disappearing from another copy of it in mid-test?

In the end I decided to switch to just replacing also unsued `solc` modules I expect to be unused with the one from the tarball. It should work just as well for making sure we don't have old solc-js there but hopefully with less breakage. The downside is only that each installation of the tarball adds some extra time to CI, especially for Truffle because it seems to recompile its TS code on each some operation.

Now I'm going to reorganize my and your branches for the merge so please don't push anything in the meantime. I'll close this PS and put your two other PRs at the bottom because they pass tests on `master` and we can merge them without special intervention from @axic. I'll put mine on top of that with the changes you added here.

cameel:
#594, #595 and #596 are now ready and this PR has served its purpose so I'm closing it.

PR #600, opened 2022-02-12
Refactor and provide type support to linker.ts 

stephensli:
# Why

This change provides some basic type support for the internal and exported methods within `linker.ts`, includes support documentation where missing for the consumer. The target result of this `PR` is to increase readability, maintainability and start the introduction of types.

This is the continuous development happening on top of #566 to improve user consumption, and development experience. 

Depends on #594

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,5 @@ out
 dist/**
 
 .nyc_output
+.idea
+.DS_Store
--- a/common/helpers.ts
+++ b/common/helpers.ts
@@ -1,0 +1,24 @@
+/**
+ * returns true if and only if the value is null or undefined.
+ * Uses == over === which compares both null and undefined.
+ * @param value
+ */
+function isNil (value: any): boolean {
+  return value == null;
+}
+
+/**
+ * returns true if and only if the value is a object and not an array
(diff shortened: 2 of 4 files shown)


axic (changes requested):
This PR mixes changes (refactoring) and type support. I'd prefer to minimize the mixing of changes as it makes it harder to review and verify that nothing went awry.

stephensli:
@axic I find the PR break down here a little extreme, i'm happy to make the break up but seems to cause more headaches than contribute. This PR will be changed to contain only the refactor.

stephensli:
Refactor only PR: #603

stephensli:
Types only PR: #604

cameel:
> I find the PR break down here a little extreme, i'm happy to make the break up but seems to cause more headaches than contribute. This PR will be changed to contain only the refactor.

I find that just a good practice to make independent things independent (though a lot of the time just having separate commits is good enough for me). But apart from that the situation here is also a bit special because we're actually bottlenecked by reviews. So optimizing for readability and making smaller PRs without mixing stuff together (even if there's more of them) is what will help to push these changes through faster.

PR #623, opened 2022-05-17
Add the missing `common/helpers.ts` to `package.json`

### LINKED ISSUE #622: `Uncaught Error: Cannot find module './common/helpers'` on 0.8.14
Looks like the helpers added in #603 were not added to the package so trying to import it fails:
```bash
npm install solc@0.8.14
node_modules/solc/solc.js --version
```
```
Error: Cannot find module './common/helpers'
```
Also:
```bash
node -e 'require("solc")'
```
```
Error: Cannot find module './common/helpers'
```

Also, we need to investigate why this didn't fail any tests. We specifically build the package in CI and install it and that test should have caught it.

ekpyron:
Fixes #622.

I guess this is enough to fix the issue, but maybe better if @cameel and/or @stephensli confirmed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "solc.js",
     "soljson.js",
     "translate.js",
-    "wrapper.js"
+    "wrapper.js",
+    "common/helpers.js"
   ],
   "author": "chriseth",
   "license": "MIT",


stephensli:
@cameel this will do it, please just do a npm build-tarball and check the tarball output to see if the content contains everything required. If anything else is missing (hopefully not) include it in this PR @ekpyron

chriseth (approved):

nataouze:
Looks like "common/types.js" should be added too as it is also imported in linker.ts

ekpyron:
As far as I can see ``dist/common/types.js`` (i.e. the js file built from ``common/types.ts``), is empty and not imported by any of the other built js files - probably since it's only relevant for typescript and does not carry over to js...

nataouze:
Yes you are right, it is not present in the built linker.js file. Thank you

stephensli:
@cameel I would not invest much effort as #614 updates the exported files.

cameel:
Well, we still need to find out why it went through our tests unnoticed. And then, when we fix the tests, we should rebase this PR on top of that to make sure the fix still passes them. I hope that the missing entry in `package.json` is the only problem but we really can't be sure of anything until tests are working properly.

PR #625, opened 2022-05-18
Fix solc-js injection in CI

cameel:
~Depends on #623.~ Merged.

This PR changes the mechanism for injecting solc-js as a dependency of Truffle and Hardhat, introduced in #594, to make it more robust.

Normally when you install a boilerplate project you get multiple copies of `solc` in the dependency tree and looks like the method I used did not actually replace all of them. The way it was done is generally very hacky - going to the dependency dir and using npm or yarn to install a new version of the dependency. It works most of the time but sometimes also changes the dependency tree in weird ways and, honestly, I don't 100% understand what npm/yarn is exactly doing. Their dependency trees are just too complex.

Now I switched to the `overrides` feature, newly added in npm 8.3. This should be more reliable. yarn has a similar feature but the key is named differently ([`resolutions`](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)) so to make things simpler, I'm just always using npm.

 I'm also adding a sanity check to make sure that there are no other versions of `solc` in the dependency tree.

When we ensure that tests do break here, I can rebase it on #623 to make sure it actually fixes the problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,20 +83,34 @@ commands:
       tarball-path:
         type: string
         default: solc-js.tgz
-      install-command:
-        type: string
-        default: npm install
+      package-manager:
+        type: enum
+        enum: ["npm", "yarn"]
+        default: npm
     steps:
+      - run:
+          name: "Sanity check: tarball exists and the target dir contains a JS project"
+          command: |
+            [[ -f "<<parameters.tarball-path>>" ]]
+            [[ -f "<<parameters.path>>/package.json" ]]
       - run:
           name: Inject solc-js from the tarball into dependencies at <<parameters.path>>
           command: |
-            [[ -f "<<parameters.tarball-path>>" ]]
             absolute_tarball_path=$(realpath "<<parameters.tarball-path>>")
-            for solc_module in $(find "<<parameters.path>>" -type d -path "*/node_modules/solc"); do
-                pushd "${solc_module}/../.."
-                <<parameters.install-command>> "$absolute_tarball_path" --ignore-workspace-root-check
-                popd
-            done
+            cd "<<parameters.path>>"
+            mv package.json original-package.json
+            # NOTE: The 'overrides' feature requires npm >= 8.3. Yarn requires `resolutions` instead.
+            jq ". + {overrides: {solc: \"${absolute_tarball_path}\"}} + {resolutions: {solc: \"${absolute_tarball_path}\"}}" original-package.json > package.json
+            "<<parameters.package-manager>>" install
+      - run:
+          name: "Sanity check: all transitive dependencies successfully replaced with the tarball"
+          command: |
+            solc_version=$(jq --raw-output .version solc-js/package.json)
+            cd "<<parameters.path>>"
+            if "<<parameters.package-manager>>" list --pattern solc | grep 'solc@' | grep -v "solc@${solc_version}"; then
+              echo "Another version of solc-js is still present in the dependency tree."
+              exit 1
+            fi
 
   provision-and-package-solcjs:
     description: "Creates a package out of latest solc-js to test its installation as a dependency."
@@ -124,7 +138,7 @@ commands:
           dependency-file: yarn.lock
       - inject-solc-js-tarball:
           path: hardhat/
-          install-command: yarn add
+          package-manager: yarn
 
   provision-truffle-with-packaged-solcjs:
     description: "Clones Truffle repository and configures it to use a local clone of solc-js."
@@ -132,18 +146,8 @@ commands:
       - run: git clone --depth 1 "https://github.com/trufflesuite/truffle" truffle/
       - install-truffle-dependencies
       - inject-solc-js-tarball:
-          path: truffle/node_modules/
-          install-command: yarn add
-      - run:
-          name: Neutralize any copies of solc-js outside of node_modules/
-          command: |
-            # NOTE: Injecting solc-js into node_modules/ dirs located under truffle/packages/ causes
-            # an error 'Tarball is not in network and can not be located in cache'. These are not
-            # supposed to be used but let's remove them just in case.
-            find truffle/ \
-              -path "*/solc/wrapper.js" \
-              -not -path "truffle/node_modules/*" \
-              -printf "%h\n" | xargs --verbose rm -r
+          path: truffle/
+          package-manager: yarn
 
 jobs:
   node-base: &node-base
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -76,20 +83,27 @@ commands:
...
+            mv package.json original-package.json
+            # NOTE: The 'overrides' feature requires npm >= 8.3
+            jq ". + {overrides: {solc: \"${absolute_tarball_path}\"}}" original-package.json > package.json
+            npm install
+      - run:
+          name: "Sanity check: all transitive dependencies successfully replaced with the tarball"
+          command: |
+            solc_version=$(jq --raw-output .version solc-js/package.json)
+            cd "<<parameters.path>>"
+            npm list solc | grep 'solc@' | grep -v "solc@${solc_version}"
```
cameel, line 106: Doing this with grep is not great but a more reliable check will be much longer. `npm list` can give me JSON but it's deeply nested. I tried to create a `jq` command that would handle it but it's getting complicated so I decided to stay with grep for now.

.circleci/config.yml
```diff
@@ -277,6 +277,8 @@ jobs:
   truffle-sample-project:
     docker:
       - image: circleci/node:12
+    # npm install runs out of memory on medium when installing Truffle's dependencies
```
ekpyron, line 280: Wtf... how much memory is that :-D? Kind of crazy :-).

cameel: Truffle's build process is actually very heavy for a JS app. Typescript compilation takes several minutes and that's probably where the memory goes. I don't really want to go too deep into it so for now I'm just increasing the container size.

But I'm also switching to yarn right now since the dependency injection does not work well with npm for Truffle. Maybe that eats less, though if it's because of TS then probably not.

I already had to deal with that a bit when I was creating this config for the first time. At least in terms of build time (did not have any memory issues on `medium` back then). The CI commands are structured in such a way as to avoid rerunning `yarn install` for Truffle if not necessary because that rebuilds everything.

.circleci/config.yml
```diff
@@ -284,6 +301,35 @@ jobs:
...
+          command: |
+            mkdir package/
+            cd package/
+            npm install ../solc-js.tgz
+
+            npx solcjs --version
+
+            echo "contract C {}" > C.sol
+            npx solcjs C.sol --bin
+            [[ -f C_sol_C.bin ]]
```
cameel, line 331: This is a new test, not completely related to the PR but it's also a case that fails with the same error.

I could make this a separate PR but we should merge them together anyway.

ekpyron: Fine to keep here - I actually thought about suggesting to add something simple like this :-).

.circleci/config.yml
```diff
@@ -76,20 +84,34 @@ commands:
...
-            for solc_module in $(find "<<parameters.path>>" -type d -path "*/node_modules/solc"); do
-                pushd "${solc_module}/../.."
-                <<parameters.install-command>> "$absolute_tarball_path" --ignore-workspace-root-check
-                popd
-            done
+            cd "<<parameters.path>>"
+            mv package.json original-package.json
+            # NOTE: The 'overrides' feature requires npm >= 8.3. Yarn requires `resolutions` instead.
+            jq ". + {overrides: {solc: \"${absolute_tarball_path}\"}} + {resolutions: {solc: \"${absolute_tarball_path}\"}}" original-package.json > package.json
+            "<<parameters.package-manager>>" install
```
cameel, line 105: Doing `yarn install` here increases the running time of the Truffle test a lot. That's because it recompiles all TypeScript files and that's really heavy. `install-truffle-dependencies` at this point already ran `yarn install` so we just redo that work.

The problem is that it's hard to avoid it while still keeping it as a separate, self-contained command. I'd have to duplicate this stuff in `install-truffle-dependencies` but still keep it here for Hardhat. Maybe I could figure out something nicer but it's a refactor that I don't really want to go into right now.

.circleci/config.yml
```diff
@@ -172,11 +185,23 @@ jobs:
...
+          name: Restore the default solc binary expected by Hardhat
+          command: |
+            # Hardhat downloader tests are hard-coded to expect the version that comes with the solc-js.
+            # We forced latest solc-js but we still want the default binary with it.
+            hardhat_default_solc_version=$(jq --raw-output '.dependencies.solc' hardhat/packages/hardhat-core/package.json)
+            mkdir hardhat-default-solc/
+            pushd hardhat-default-solc/
+            npm install "solc@${hardhat_default_solc_version}"
+            popd
+            ln -sf ../../../hardhat-default-solc/node_modules/solc/soljson.js hardhat/node_modules/solc/soljson.js
```
cameel, line 202: This part is needed for the downloader tests to pass. Otherwise when the original import problem is resolved they still fail like in [5174](https://app.circleci.com/pipelines/github/ethereum/solc-js/1175/workflows/1203dc75-62d7-435b-b66b-04008e693202/jobs/5174):

```
  1) Compiler downloader
       Downloaded compiler verification
         Shouldn't do anything if the compiler is fine:
     Error: ENOENT: no such file or directory, unlink '/tmp/hardhat-tests-compiler-downloader/wasm/list.json'
  

  2) Compiler downloader
       getDownloadedCompilerPath
         If the compiler already existed
           Should return it if it's passes the verification:
     Error: ENOENT: no such file or directory, stat '/home/circleci/project/hardhat/node_modules/solc/soljson.js'
```

The reason seems to be that while Hardhat ignores the binary from the package at runtime, [it does expect it to be there in tests](https://github.com/NomicFoundation/hardhat/blob/hardhat%402.9.5/packages/hardhat-core/test/internal/solidity/compiler/downloader.ts#L59). And it [assumes it to be specifically 0.7.3](https://github.com/NomicFoundation/hardhat/blob/hardhat%402.9.5/packages/hardhat-core/test/internal/solidity/compiler/downloader.ts#L29-L41). I think that the first test that fails tries to verify the binary and verification routine is deletes it if it's not the version it expects, which is the case here.

I think that the only reason it wasn't failing before was that the solc-js injection into dependencies was not being performed correctly.

ekpyron: Given all this craziness I think that the simple test in https://github.com/ethereum/solc-js/pull/625#discussion_r876059739 will probably be the most reliable indicator in the future :-).

cameel:
Tests passed on #623 and here they now pass for the right reason (i.e. the missing `common/helpers.ts`) so I'm rebasing it on #623.

Failing tests without #623: [d866014e-000b-4897-9eb7-c8d88ad1a5fe](https://app.circleci.com/pipelines/github/ethereum/solc-js/1176/workflows/d866014e-000b-4897-9eb7-c8d88ad1a5fe)
Passing tests with #623: [04470bb5-dd5b-4211-8a7d-88607023b9b5](https://app.circleci.com/pipelines/github/ethereum/solc-js/1178/workflows/04470bb5-dd5b-4211-8a7d-88607023b9b5)

.circleci/config.yml
cameel: Yeah, for simple crashes like this one. For more tricky cases without an outright crash this one might still be useful.

ekpyron (approved):
We did verify that this actually works, right :-)? Then I'd say, we merge.

cameel:
Yeah. See my comment above for CI runs showing that it would have failed before our earlier fix and is passing after it.

PR #626, opened 2022-05-19
Fix CLI `--version` test to use version from the binary not `package.json`

cameel:
Currently tests in `cli.ts` are hard-coded to require a binary with version matching `package.json`. This not true in general and causes problems in #598.

This PR changes the test to only require the output from `--version` to match the output of `version()` function in the binary. I also kept a modified version of the previous check as a sanity check - to match that the output actually looks like our version string.

### DIFF
--- a/test/cli.ts
+++ b/test/cli.ts
@@ -1,12 +1,13 @@
 import tape from 'tape';
 import spawn from 'tape-spawn';
 import * as path from 'path';
-const pkg = require('../package.json');
+import solc from '../';
 
 tape('CLI', function (t) {
   t.test('--version', function (st) {
     const spt = spawn(st, './solc.js --version');
-    spt.stdout.match(RegExp(pkg.version + '(-[^a-zA-A0-9.+]+)?(\\+[^a-zA-Z0-9.-]+)?'));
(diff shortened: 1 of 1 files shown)


test/cli.ts
```diff
@@ -1,12 +1,13 @@
 import tape from 'tape';
 import spawn from 'tape-spawn';
 import * as path from 'path';
-const pkg = require('../package.json');
+import solc from '../';
 
 tape('CLI', function (t) {
   t.test('--version', function (st) {
     const spt = spawn(st, './solc.js --version');
-    spt.stdout.match(RegExp(pkg.version + '(-[^a-zA-A0-9.+]+)?(\\+[^a-zA-Z0-9.-]+)?'));
```
cameel, line 9: This old regex actually seems very broken. Given that the string it's meant to match is something like `0.8.10+commit.fc410830.Emscripten.clang`:
- `A-A` should probably have been `A-Z`.
- The character groups should not be negated with `^`.

I think that only reason this regex "worked" is that these parts are optional (`?`) and the regex was not anchored with `^` and `$`.

ekpyron (approved):

PR #630, opened 2022-06-02
Add types for compiler input & output

GrandSchtroumpf:
This Pull Request adds types & inline documentation for compiler input & output based on the [Compiler Input & Output documentation](https://docs.soliditylang.org/en/v0.8.14/using-the-compiler.html#compiler-input-and-output-json-description) and the [ABI Spec documentation](https://docs.soliditylang.org/en/v0.8.14/abi-spec.html).
Some types are not strictly typed (Yul AST for example) because I didn't find the right documentation. Don't hesitate to point me to some doc / code to improve it.

**Note**: I included a commit that demonstrate how to use the output type with the update method. I can remove the commit from the PR for separation of concern.

### DIFF
--- a/common/compiler-input.ts
+++ b/common/compiler-input.ts
@@ -1,0 +1,261 @@
+/***********/
+// SOURCES //
+/***********/
+
+export interface SourceInputUrls {
+  /** Hash of the source file. It is used to verify the retrieved content imported via URLs */
+  keccak256?: string
+  /**
+   * URL(s) to the source file.
+   * URL(s) should be imported in this order and the result checked against the
+   * keccak256 hash (if available). If the hash doesn't match or none of the
+   * URL(s) result in success, an error should be raised.
+   */
+  urls: string[]
+}
+export interface SourceInputContent {
+  /** Hash of the source file. */
+  keccak256?: string
+  /** Literal contents of the source file */
+  content: string
+}
+
+export interface SourcesInput {
+  [contractName: string]: SourceInputContent | SourceInputUrls
+}
+
+/********************/
+// OUTPUT SELECTION //
+/********************/
+export type OutputType =
+  | 'abi'
+  | 'devdoc'
+  | 'userdoc'
+  | 'metadata'
+  | 'ir'
+  | 'irOptimized'
+  | 'storageLayout'
+  | 'evm.assembly'
+  | 'evm.legacyAssembly'
+  | 'evm.bytecode.functionDebugData'
+  | 'evm.bytecode.object'
+  | 'evm.bytecode.opcodes'
+  | 'evm.bytecode.sourceMap'
+  | 'evm.bytecode.linkReferences'
+  | 'evm.bytecode.generatedSources'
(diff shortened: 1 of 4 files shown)


common/compiler-input.ts
```diff
@@ -0,0 +1,261 @@
...
+
+export interface CompilationInput {
+  /** Source code language */
+  language: 'Solidity' | 'yul'
+  sources: SourcesInput
+  settings?: CompilerSettings
+}
+
+export interface CondensedCompilationInput {
+  language: 'Solidity' | 'Vyper' | 'lll' | 'assembly' | 'yul'
```
cameel, line 256: What is this based on? Technically, [solc supports only `Solidity` and `Yul` (case-sensitive)](https://github.com/ethereum/solidity/blob/a53f15f45f9089623ec91961602f6a07ff0f086e/libsolidity/interface/StandardCompiler.cpp#L1514-L1519). In a broader context that also includes any compilers supporting Standard JSON I'd say that the values are not limited to some predefined choices - other compilers are free to support any language identifiers.

common/compiler-input.ts
```diff
@@ -0,0 +1,261 @@
...
+   * - evm.gasEstimates - Function gas estimates
+   * - ewasm.wast - Ewasm in WebAssembly S-expressions format
+   * - ewasm.wasm - Ewasm in WebAssembly binary format
+   */
+  outputSelection?: CompilerOutputSelection
+  /** The modelChecker object is experimental and subject to changes. */
+  modelChecker?: CompilerModelChecker
+}
+
+export interface CompilationInput {
```
cameel, line 248: These definitions are pretty detailed. There are some problems though:
1) If we keep them here, they will be hard to keep in sync with the compiler. Maybe there are better ways to go about this? For example defining some JSON schema (not tied specifically to TypeScript syntax) that would be kept in the main compiler repo and only imported here?
2) solc-js is meant to support all the old compiler versions and Standard JSON changed over time. So these definitions won't be right for all the compiler binaries that you can use. So we either need separate definitions for different versions (at least breaking ones) or some superset that won't exactly fit any particular version.

cameel (commented):
There are some issues we should discuss first before we dive into actually reviewing it in detail. Please see my comments below.

@axic, curious what's your opinion on this. Is it even feasible to have a precise schema for Standard JSON in solc-js given that it's not consistent between compiler versions? This kinda gets into the territory of https://github.com/ethereum/solidity/issues/2276.

GrandSchtroumpf:
@cameel thanks for the review. I created these type for `remix-plugin` two years ago, I realized that some types are not well defined, I'll update it shortly.
We can definitely have a script generating the types based on a version. I think this would be the best long term solution.

I think this PR could still be a good place to discuss type naming, how to deal with the different output based on the input config, ... like that we can have a base for generating types on a second step. What do you think ?

GrandSchtroumpf:
Ok no problem. I can continue the output types on an external library for now, and if you consider it would make sense to add it into solc-js at some point I can make another PR 🙂

PR #633, opened 2022-06-10
Switch away from deprecated `circleci/node` images and update node.js in CI to 18

cameel:
Related to https://github.com/ethereum/solidity/pull/13068 and the [feeback we received in CircleCI's config review](https://gist.github.com/jenny-miggin/9578d2e52b383b3787e265c80e5f6b03).

Since the `circleci/` images are no longer updated there's no version 18 so this is needed if we want to use any newer versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -309,8 +309,8 @@ jobs:
           command: |
             cd metacoin/
             # `truffle test` compiles the project but artifacts go into /tmp/
-            ! [[ -e build/ ]] || false
-            echo "module.exports['compilers'] = {solc: {version: '$(realpath node_modules/solc/)'}}" > truffle-config.js
+            ! [[ -e build/ ]]
+            echo "module.exports['compilers'] = {solc: {version: '$(realpath ../truffle/node_modules/solc/)'}}" >> truffle-config.js
             node ../truffle/node_modules/.bin/truffle test
 
   cli-smoke-test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ workflows:
       - node-v14
       - node-v16:
           run_coveralls: true
-      - node-v17
+      - node-v18
       - hardhat-core-default-solc
       - hardhat-core-latest-solc
       - hardhat-sample-project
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ commands:
       - run:
           name: Update globally available npm to the latest version
           # Note: We need npm >= 8.3 which supports 'overrides' in package.json
-          command: sudo npm install npm --global
+          command: npm install npm --global
```
cameel, line 31: In the new `cimg/` images this not only does not require `sudo` but actually does not work at all with it:
```
sudo: npm: command not found
```

.circleci/config.yml
```diff
@@ -183,9 +183,8 @@ jobs:
 
   hardhat-core-default-solc:
     docker:
-      - image: circleci/node:16
+      - image: cimg/node:16.15
     steps:
-      - update-npm
```
cameel, line 188: `cimg/node:16` comes with npm 8.11.0 so this is no longer needed. So we only need it for the one Truffle test that requires node.js 10 and I hope that we'll be able to update that one soon too.

Maybe we should replace this with a check for min npm version but I'm not sure if it's worth the hassle since this will become irrelevant soon.

.circleci/config.yml
```diff
@@ -314,9 +311,8 @@ jobs:
 
   cli-smoke-test:
     docker:
-      - image: circleci/node:17
+      - image: cimg/node:current
```
cameel, line 314: Not sure why I had this one set specifically to 17. It should have been just latest version.

ekpyron (approved):

cameel:
Looks like the new `node` images do not have Python even though node-gyp depeds on it. Adding.

.circleci/config.yml
```diff
@@ -284,11 +281,13 @@ jobs:
...
   truffle-sample-project:
     docker:
-      - image: circleci/node:16
+      - image: cimg/node:16.15
     steps:
       - update-npm
       - show-npm-version
       - provision-and-package-solcjs
+      - run: sudo apt update
+      - run: sudo apt install python3 --assume-yes --no-install-recommends
```
ekpyron, line 290:
```suggestion
      - run: sudo apt install python3 python-is-python3 --assume-yes --no-install-recommends
```
might fix the sqlite thing I see in CI?

cameel: That worked. Thanks.

I also had to do another fix - Truffle updated its `metacoin` box and it seems to work a bit differently. A small config tweak solved it though. Now the job passes.

PR #640, opened 2022-07-08
Adding command to copy meta files into dist dir

### LINKED ISSUE #635: README and LICENSE are missing from the npm package
Until [0.8.11](https://www.npmjs.com/package/solc/v/0.8.11) the package page on npm had our README. In [0.8.12](https://www.npmjs.com/package/solc/v/0.8.12) README is no longer there. LICENSE file is not there either.

I think they disappeared because we now package only files from the `dist/` subdir. Any extra files need to be explicitly copied there.

We need go add these two back. Do we need any other files besides them?

derrickpersson:
Fixes #635.
Closes #647.

Adding the LICENSE and README.md files back to tarball.

Output from the `npm run build:tarball` command:
Screenshot 2022-07-07 at 11 29 55 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "postbuild": "node build/postbuild.js",
+    "postbuild": "node build/postbuild.js && cp README.md LICENSE dist/",
     "lint": "eslint --ext .js,.ts .",
     "lint:fix": "eslint --fix --ext .js,.ts .",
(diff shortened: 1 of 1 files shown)


derrickpersson:
I don't believe the failure in the truffle-sample-project is related to my change.
Looks like a dependency may have been updated.

cameel (changes requested):

cameel (approved):
I pushed the changes from #647 here and also rebased this on master after merging #633.

This one is now good so I'm merging it too.

PR #643, opened 2022-07-15
Changed Error in cli.ts from SyntaxError to just Error

nishant-sachdeva:
This change is in line with the change in solidity repo. Error with type information are replaced with severity values. So, only the word Error appears

Made to fix CI errors appearing in https://github.com/ethereum/solidity/pull/13162


cameel:
Closing since this needs a different solution. The compiler output should remain unchanged after the refactor from https://github.com/ethereum/solidity/pull/13162 and therefore no adjustments in solc-js should be required.

Issue #632, opened 2022-06-10
Test that `downloadCurrentVersion.ts` actually downloads the latest version


cameel:
We used to test that indirectly be checking that `solcjs --version` returns a version matching the one from `package.json`. I changed that in #626 because this is not really guaranteed - the returned version depends on the binary. So now we should also add a proper test that `downloadCurrentVersion.ts` actually does it job and downloads the right binary. It could be a JS test or and additional check in one of the jobs in CI.

OrsiniBr:
Is this issue still open.. I'd love to solve

PR #648, opened 2022-08-07
Add a test to check the download of the latest version

r0qs:
This PR attempt to solve the issue https://github.com/ethereum/solc-js/issues/632. It adds a test to verify if the current version is equal to the downloaded one, and it also modifies the downloader script to enable its use in the tests.


cameel:
Thanks for the PR!

I'll take a closer look at it later but for now just wanted to say that `solidity-solcjs-ext-test` job is failing so there must still be something wrong there.

Also, the coverage bot says the coverage decreased, which means you need to add some more tests.

r0qs:
The coverage decrease seems to be because I separated the `downloadCurrentVersion.ts` into two files, exported `getVersionList` and `downloadBinary` to be used on the test, but didn't add tests in case the download fails (e.g., HTTP 404).
I believe these were not covered before either because the original code only logs and exits on error.

But I can add tests for it. However, maybe it would be better to mock the HTTP requests in those methods and pass the endpoint URL as parameter. What do you think?

About the CI, the problem is that I'm getting the latest version based on the `version` field in the `package.json`.
But the CI updates the version to one minor version ahead, which was not released yet, so the test fails:
```
/tmp/ext-test-solc-js-nIDFB9
Using compiler version 0.8.16-ci.2022.8.5+commit.49a2db99.Emscripten.clang
Updating index.js file...
Copying determinism.js...
Copying contracts...
Copying SMTChecker tests...
Updating package.json to version 0.8.16
v0.8.16
Replacing fixed-version pragmas...
Running test function...
...

not ok 1120 Version 0.8.16 is not the latest release 0.8.15

...

```

Maybe I misunderstood the goal of the task. Should I not use the latest version in the `package.json` when trying to download? And instead, should I only test if the latest available version in the endpoint is the one that got downloaded?

cameel:
> But I can add tests for it. However, maybe it would be better to mock the HTTP requests in those methods and pass the endpoint URL as parameter. What do you think?

Yeah, definitely. I actually think that in the JS tests we would be better off with mocks both for positive and negative case. I'd consider an actual download an integration test and for that we might just go all the way - i.e. have a CI job that executes the script and does the check in Bash rather than trying to do it in JS. But, well, I'm not against having that part in JS - it's not like other tests here are pure unit tests :) It's a bit of a mess so whatever works is fine.

BTW, would mocking require adding some dependencies? We overall prefer to avoid too many deps - we have enough already and we should actually trim that even more.

> Maybe I misunderstood the goal of the task. Should I not use the latest version in the `package.json` when trying to download?

Are you sure you're really getting the latest **release** version? `0.8.16-ci.2022.8.5+commit.49a2db99.Emscripten.clang` is not a release version. It's a build from CircleCI. Not sure why you're actually getting this - https://binaries.soliditylang.org/bin/ (https://github.com/ethereum/solc-bin/tree/gh-pages/bin) has nightlies too and you might be getting that but then you'd have `nightly` in the version string, not `ci`.

r0qs:
> Yeah, definitely. I actually think that in the JS tests we would be better off with mocks both for positive and negative case. I'd consider an actual download an integration test and for that we might just go all the way - i.e. have a CI job that executes the script and does the check in Bash rather than trying to do it in JS. But, well, I'm not against having that part in JS - it's not like other tests here are pure unit tests :) It's a bit of a mess so whatever works is fine.

Great. I will do that then. Mocking the JS tests and adding the integration test in the CI.

> BTW, would mocking require adding some dependencies? We overall prefer to avoid too many deps - we have enough already and we should actually trim that even more.

No, not necessarily. We can use the native HTTPS module of Node.js.

> Are you sure you're really getting the latest **release** version? `0.8.16-ci.2022.8.5+commit.49a2db99.Emscripten.clang` is not a release version. It's a build from CircleCI. Not sure why you're actually getting this - https://binaries.soliditylang.org/bin/ (https://github.com/ethereum/solc-bin/tree/gh-pages/bin) has nightlies too and you might be getting that but then you'd have `nightly` in the version string, not `ci`.

I'm getting the current version from the package.json. See [this](https://github.com/ethereum/solc-js/pull/648/files#diff-578495d3f753e0d5300d398ad6f8f2059c1b706386e12cdec8fc3285d23d2aceR13) line. And the released version is fetched from the "releases" entry in the list.json in [this](https://github.com/ethereum/solc-js/pull/648/files#diff-578495d3f753e0d5300d398ad6f8f2059c1b706386e12cdec8fc3285d23d2aceR18) line.
The problem was that version `0.8.16` was set in the package.json by the CI but did not exist in the list.json as a release.

downloader.ts
```diff
@@ -0,0 +1,64 @@
+import * as fs from 'fs';
+import { https } from 'follow-redirects';
+import MemoryStream from 'memorystream';
+import { keccak256 } from 'js-sha3';
+
+function getVersionList () {
+  console.log('Retrieving available version list...');
+
+  return new Promise<string>((resolve, reject) => {
```
cameel, line 9: Any reason to directly use promises rather than `async`/`await` here?

downloader.ts
```diff
@@ -0,0 +1,64 @@
...
+    const mem = new MemoryStream(null, { readable: false });
+    https.get('https://binaries.soliditylang.org/bin/list.json', function (response) {
+      if (response.statusCode !== 200) {
```
cameel, line 12: It's something that has already been here before but we should change this so that anything between 200 and 299 is a success. I think we also have the same check in other places in the code

But I'd do it in a separate PR since it's a functional change, not just a refactor and is also unrelated to tests.

downloadCurrentVersion.ts
```diff
@@ -3,73 +3,26 @@
...
-  if (fs.existsSync(outputName)) {
-    fs.unlinkSync(outputName);
-  }
-
-  process.on('SIGINT', function () {
-    console.log('Interrupted, removing file.');
-    fs.unlinkSync(outputName);
+    const expectedHash = expectedFile.keccak256;
+    await downloader.downloadBinary('soljson.js', releaseFileName, expectedHash);
+    process.exit();
```
cameel, line 20: Do we need an explicit exit? Wouldn't it be better to let the script simply end?

test/downloader.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (semver.neq(solc.version(), wanted)) {
+        st.fail('Downloaded version differs from package version');
+      }
+      st.pass(`Version ${wanted} successfully downloaded`);
+    } catch (err) {
+      st.fail(err.message);
+    }
+    st.end();
+  });
+});
```
cameel, line 38: I don't think this test is actually testing what it should. What we need is a test that `downloadCurrentVersion.ts` will download the version we have in `package.json` but this test does not currently even import that script. Instead it uses the `downloader`, which does not have the logic for selecting the binary. It just gets the whole list and gets whichever binary you tell it to get.

As discussed in https://github.com/ethereum/solc-js/pull/648#issuecomment-1210915634, I think we should have a general integration test invoking the whole script at Bash level. Then this JS test could be replaced with some unit tests (with mocks) for `downloadBinary()` and `getVersionList()` instead.

cameel:
> I'm getting the current version from the package.json. See [this](https://github.com/ethereum/solc-js/pull/648/files#diff-578495d3f753e0d5300d398ad6f8f2059c1b706386e12cdec8fc3285d23d2aceR13) line. And the released version is fetched from the "releases" entry in the list.json in [this](https://github.com/ethereum/solc-js/pull/648/files#diff-578495d3f753e0d5300d398ad6f8f2059c1b706386e12cdec8fc3285d23d2aceR18) line.
> The problem was that version `0.8.16` was set in the package.json by the CI but did not exist in the list.json as a release.

Oh, you're right. Looks like the [external test for solc-js that we have in the main repo actually bumps version in `package.json`](https://github.com/ethereum/solidity/blob/develop/test/externalTests/solc-js/solc-js.sh#L57-L59). It's actually that test that's getting a binary from CI - and it's because it's supposed to test that the latest compiler binary works with solc-js. It's bumping version because solc-js and the compiler have different conventions for updating the version. In the main repo, we bump [the version hard-coded in the main `CMakeLists.txt`](https://github.com/ethereum/solidity/blob/develop/CMakeLists.txt#L24) immediately after a release while in solc-js we bump version in `package.json` just before a release. This means that version in solc-js is one number behind solc for most of the development cycle.

In any case, this just shows that this must be tested from the outside. I.e. we should not be assuming inside these JS tests that version in `package.json` is the most recent one. Instead this should be a Bash integration that installs solc-js on its own (so it can ensure that `package.json` was not changed), invokes `downloadCurrentVersion.ts` and makes sure it downloaded the right thing.

downloader.ts
r0qs: Yes. As far as I know the `https.get`module from nodejs is not "awaitable" so I needed to wrap it in a promise to explicitly resolve the response.

downloadCurrentVersion.ts
r0qs: No we don't. I'll remove it.

r0qs:
I added the unit tests for `downloadBinary` and `getVersionList`. I decided to use a dependency called `nock` to easily mock the HTTPS requests.
I also added a dummy certificate that is only used in the tests, otherwise we would need to rewrite the `downloadBinary` and `getVersionList` to not use the HTTPS module directly but also support HTTP.

I opened another PR https://github.com/ethereum/solc-js/pull/659 for the CI bash integration test.

test/downloader.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function generateTestFile (t: tape.Test, content: string): tmp.FileResult {
+  // As the `keep` option is set to true the removeCallback must be called by the caller
+  // to cleanup the files after the test.
+  const file = tmp.fileSync({ template: 'soljson-XXXXXX.js', keep: true });
+  try {
+    fs.writeFileSync(file.name, content);
+  } catch (err) {
+    t.fail('error writing test file');
```
matheusaaguiar, line 29: I think that would be nice to know why the file could not be written, i.e, was lack of permission, lack of memory or something else?

matheusaaguiar (commented):
I am actually learning JS/TS more than anything with this review, but here's my 2 cents :)

downloadCurrentVersion.ts
```diff
@@ -3,73 +3,27 @@
...
+    const releaseFileName = list.releases[version];
+    const expectedFile = list.builds.filter(function (entry) { return entry.path === releaseFileName; })[0];
 
-  const mem = new MemoryStream(null, { readable: false });
-  https.get('https://binaries.soliditylang.org/bin/list.json', function (response) {
-    if (response.statusCode !== 200) {
-      console.log('Error downloading file: ' + response.statusCode);
-      process.exit(1);
+    if (!expectedFile) {
+      throw new Error('Requested version not found. Version list is invalid or corrupted?');
```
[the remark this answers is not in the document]
r0qs: Yes, but in this case it would not find the version in the release list and throw the same error.

test/downloader.ts
```diff
@@ -0,0 +1,190 @@
...
+
+const assets = path.resolve(__dirname, 'resources/assets');
+
+tape.onFinish(() => {
+  if (!nock.isDone()) {
+    throw Error('expected requests were not performed');
+  }
+});
+
+function hash (filePath: string): string {
```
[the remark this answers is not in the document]
r0qs: Yes, indeed. Maybe it should also be renamed to `hashFile` or something like that? Since it reads and hashes the content of a file.

test/downloader.ts
r0qs: Yes, I will add the `err` in the error message.

test/downloader.ts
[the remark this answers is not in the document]
r0qs: Actually, I think we should be move to `common/helpers.ts` since it is not related with the download logic.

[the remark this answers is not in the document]
matheusaaguiar: Yeah, I guess that's better.

downloadCurrentVersion.ts
```diff
@@ -3,73 +3,27 @@
...
 const pkg = require('./package.json');
+const DEFAULT_HOST = 'https://binaries.soliditylang.org';
 
-function getVersionList (cb) {
-  console.log('Retrieving available version list...');
+async function download (version, host = DEFAULT_HOST) {
+  try {
+    const list = JSON.parse(await downloader.getVersionList(host));
+    const releaseFileName = list.releases[version];
+    const expectedFile = list.builds.filter(function (entry) { return entry.path === releaseFileName; })[0];
```
nikola-matic, line 14: I'm assuming `list.builds` is an array since you're calling `filter` on it - would it make sense to use `find` instead?

downloader.ts
nikola-matic: `response.ok`?

downloader.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  return new Promise<void>((resolve, reject) => {
+    // Remove if existing
+    if (fs.existsSync(outputName)) {
+      fs.unlinkSync(outputName);
+    }
+
+    process.on('SIGINT', function () {
+      fs.unlinkSync(outputName);
+      reject(new Error('Interrupted, removing file.'));
```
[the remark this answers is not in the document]
r0qs: It may be. This only sets a listener to `SIGINT` signal (e.g. if the user hits `Ctrl+C` in the middle of the download).

downloadCurrentVersion.ts
r0qs: Yes. I replaced it.

test/downloader.ts
```diff
@@ -0,0 +1,187 @@
...
+import { https } from 'follow-redirects';
+import downloader from '../downloader';
+import { keccak256 } from 'js-sha3';
+import { hashFile } from '../common/helpers';
+
+const assets = path.resolve(__dirname, 'resources/assets');
+
+tape.onFinish(() => {
+  if (!nock.isDone()) {
+    throw Error('expected requests were not performed');
```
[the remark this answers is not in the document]
r0qs: I applied the suggestion but ignored the period at the end of the sentence since other suggestions also removed it.

matheusaaguiar (approved):
LGTM.

downloader.ts
r0qs: I don't think the `https` nodejs module has such `ok` property. I know that the [fecth API](https://developer.mozilla.org/en-US/docs/Web/API/Response) has it, but it is not what we are using here. However, we could mimic its behavior.

nikola-matic: Ah, you're right, I was actually looking at the fetch API when I made the comment.

Issue #627, opened 2022-05-21
`TypeError: Error resolving module specifier “solc/wrapper”` when trying to import solc-js in a browser


RickCastle2018:
I've added this line into my index.html file: `<script type="text/javascript" src="https://binaries.soliditylang.org/bin/soljson-v0.8.4+commit.c7e474f2.js"></script>`

So, as it's written in README I have `window.Module` variable, and I have to 'wrap' it accordingly to the instructions:

```
import * as wrapper from 'solc/wrapper';
const solc = wrapper(window.Module);
```

I added these lines into separate <script> block (placed after the first), but now I get this error in the browser console: `Uncaught TypeError: Error resolving module specifier “solc/wrapper”. Relative module specifiers must start with “./”, “../” or “/”`

Could you please tell what I was doing wrong?

P.S. There's also require() function option listed in Browser Usage. But how? Require() is only for Node.js usage as I know

cameel:
How are you installing/accessing solc-js inside the browser? This sounds a bit as if you just loaded the binary without having the library available.

We'll need a bit more details to debug this. Can you provide a more complete example to reproduce this? Which browser are you using?

RickCastle2018:
@cameel, sorry for long responce. The problem is documentation is not clear at all (for browser usage). I found that compiler should be somehow loaded to service worker (https://github.com/ethereum/remix-project/blob/614f726d94384a846a4c8cac81d5c802ff42ba12/libs/remix-solidity/src/compiler/compiler-worker.ts). But I still dont understand how... Could you please write complete guide for this?

RickCastle2018:
![image](https://user-images.githubusercontent.com/48931166/172057169-e47c8918-8e68-42ec-af26-88e31258c765.png)
Here're my attempts. I'm trying to use Web Workers, not Service Workers

r0qs:
Hi @RickCastle2018 what version of solc-js are you using?
When you opened the issue you state that were using `soljson-v0.8.4+commit.c7e474f2.js` but in your example you use `v0.4.5`.

You are correct about the `require`, but I believe this also apply to ES6 imports, since nodejs doesn’t natively support ES6 import directly as far as I know. The easy way is to bundle your modules need using [browserify](https://browserify.org/) or [webpack](https://webpack.js.org/) with babel.

Here a working example based on your code using the version `v0.8.4`. Please note that this is just an example and does not handle any errors. You should adapt it to your use case:
* index.html
```html
<!DOCTYPE html>
[...]

Issue #610, opened 2022-03-09
Why is `solc.lowlevel.compileCallback` `null` on solc 0.8.12?


Y-Zeus:
```javascript
const solc = require('solc');

if (typeof solc.lowlevel.compileSingle != 'function') {
		console.log('Low-level compileSingle interface not implemented by this compiler version.');
}
```
result:
```
TypeError: solc.lowlevel.compileCallback is not a function
```

solc version 0.8.12
why?

cameel:
These low-level functions were superseded by `compile()` and are no longer needed (https://github.com/ethereum/solidity/issues/3452). They were removed from the binary in 0.5.0 (https://github.com/ethereum/solidity/pull/5105).

The only reason they're still present under `lowlevel` is that solc-js is meant to support all compiler binaries, including the ones older than 0.5.0. The check will succeed if you use it with such a binary.

But you have a point in that it's not explained in the README. We should make it clearer.

Y-Zeus:
> 这些低级功能已被取代`compile()`并且不再需要（[ethereum/solidity#3452](https://github.com/ethereum/solidity/issues/3452)）。它们已从 0.5.0 ( [ethereum/solidity#5105](https://github.com/ethereum/solidity/pull/5105) ) 中的二进制文件中删除。
> 
> 它们仍然存在的唯一原因`lowlevel`是 solc-js 旨在支持所有编译器二进制文件，包括早于 0.5.0 的那些。如果您将其与此类二进制文件一起使用，则检查将成功。对于较新的，它们是`null`.
> 
> 但是你有一点，它没有在自述文件中解释。我们应该说得更清楚。

Y-Zeus:
thanks

cameel:
I'll keep this open. I think we should add something about it to the README.

r0qs:
Fixed by https://github.com/ethereum/solc-js/pull/658.

Issue #583, opened 2022-01-20
Better document loadRemoteVersion


PatrickAlphaC:
I'm trying to download versions of solidity using `loadRemoteVersion`, however, when I run:

```
solc.loadRemoteVersion("0.8.7", (resp) => {})
```
If I do a `console.log` where that method is defined, I get: 

https://binaries.soliditylang.org/bin/soljson-v0.8.10.js

Which is blank. However https://binaries.soliditylang.org/bin/soljson-latest.js isn't

Am I doing something wrong?

PatrickAlphaC:
Basically, I'm having a hard time installing or working with a different version. I can do it find with `solc` and `solc-select` via the command line. 

But I'd love to be able to switch versions either in the command line _or_ in my javascript. 

Is this possible? Or do I have to write code to download different versions?

PatrickAlphaC:
Or is the best way just to remove the package from node modules, change it in package-log.json, and then download the version I want?

ekpyron:
You need to pass the long version string to ``loadRemoteVersion``, i.e.
```
solc.loadRemoteVersion("v0.8.7+commit.e28d00a7", (err, solc) => { console.log(err, solc); });
```
should work.
You can extract the long version string for each version from https://binaries.soliditylang.org/bin/list.json

PatrickAlphaC:
Aha! I see! Thanks! 

Could I leave this issue open for the moment till we get a docs PR in?

Elyx0:
Yes please, keep it open 🙏  that's secret knowledge otherwise

PatrickAlphaC:
Yay open source :)

alexreyes:
I believe this issue can be closed @ekpyron

cameel:
True. @r0qs fixed it in #658. Closing.

Issue #334, opened 2019-01-21
Link to compiler with SRI


fulldecent:
Currently [in the README](https://github.com/ethereum/solc-js/blob/master/README.md#browser-usage):

> Add the version of solc you want to use into index.html:
>
> <script type="text/javascript" src="https://solc-bin.ethereum.org/bin/{{ SOLC VERSION }}.js"></script>
> (Alternatively use https://solc-bin.ethereum.org/bin/soljson-latest.js to get the latests version.)

This is not a security best practice. We should promote security best practices by providing a list of specific URLs and SRI hashes. Some people may prefer convenience over security and not use SRI, we can explain to them that they are taking on more risk that way.

r0qs:
Fixed by #658.

PR #658, opened 2022-09-12
Update documentation

r0qs:
The PR fixes the below issues adding documentation about their respective topics:

- Adds documentation recommending the use of [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) when fetching the binaries: https://github.com/ethereum/solc-js/issues/334
- Adds documentation about the necessity of using long version string in `loadRemoteVersion`: https://github.com/ethereum/solc-js/issues/583
- Adds note about backward compatibility of low-level legacy functions: https://github.com/ethereum/solc-js/issues/610
- Fix ES6 import syntax: https://github.com/ethereum/solc-js/issues/627


README.md
```diff
@@ -204,6 +204,8 @@ The low-level API is as follows:
 
 For examples how to use them, please refer to the README of the above mentioned solc-js releases.
 
+**Note**: These low-level functions still present for compatibility reasons to support all compiler binaries, but they were superseded by the `compile()` function and are no longer required in newer versions. The functions `compileSingle`, `compileMulti` and `compileCallback` are always `null` in the newer versions.
```
christianparpart, line 207: I'm no native English speaker, but using `present` here sounds somehow weird to me. Why not "These low-level functions remain available for compatibility reasons." IMHO a simpler sentence.

README.md
```diff
@@ -239,10 +241,20 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
     // An error was encountered, display and quit
   } else {
     // NOTE: Use `solcSnapshot` here with the same interface `solc` has
+    // For example:
+    const output = solcSnapshot.compile(/* ... */)
   }
 });
 ```
 
+The version should be in the long format string.
```
christianparpart, line 250: "should" or "must"?

README.md
```diff
@@ -309,12 +321,36 @@ Add the version of `solc` you want to use into `index.html`:
 <script
   type="text/javascript"
   src="https://binaries.soliditylang.org/bin/{{ SOLC VERSION }}.js"
+  integrity="sha256-{{ BASE64-ENCODED HASH OF SOLC VERSION  }}"
```
christianparpart, line 324: where is that `BASE64-ENCODED HASH OF SOLC VERSION` coming from?

README.md
```diff
@@ -309,12 +321,36 @@ Add the version of `solc` you want to use into `index.html`:
...
+  crossorigin="anonymous"
 ></script>
 ```
 
-(Alternatively use `https://binaries.soliditylang.org/bin/soljson-latest.js` to get the latests version.)
+This will load `solc` into the global variable `window.Module`.
```
christianparpart, line 329: Into the global variable or is that a namespace / object? Maybe saying assigning it as a member of `window.Module` would be more precise?

README.md
```diff
@@ -309,12 +321,36 @@ Add the version of `solc` you want to use into `index.html`:
...
+<script
+  type="text/javascript"
+  src="https://binaries.soliditylang.org/bin/soljson-v0.8.16+commit.07a7930e.js"
+  integrity="sha256-J7KCDvk4BaZcdreUWklDJYLTBv0XoomFcJpR5kA2d8I="
+  crossorigin="anonymous"
+></script>
+```
 
-This will load `solc` into the global variable `window.Module`. Then use this inside Javascript as:
+Then use this inside Javascript as:
```
christianparpart, line 353:
```suggestion
Then use this inside JavaScript as:
```
I think that's CamelCase, isn't it?

README.md
r0qs: `must`. Otherwise, it will not find the binary.

README.md
r0qs: Yes. Maybe I can put in between underscore if that makes it easier to understand that it is just a placeholder.

README.md
r0qs: I think `global variable` is correct. I mean, the `Module` is a property of window and thus a global variable since window is a global object of a web page that refers to itself. I may be wrong, but solc is not assigned as a member of the `window.Module`, it is loaded as the Module object in the window. Maybe `global object` would be correct, too. Like:
```
This will load solc as the global object window.Module.
```

But honestly, I'm not sure about the terminology. I usually refer to any window property as global variables, but I may be calling them by the wrong name, haha

README.md
r0qs: I will change it :)

README.md
```diff
@@ -308,13 +321,37 @@ Add the version of `solc` you want to use into `index.html`:
...
-  src="https://binaries.soliditylang.org/bin/{{ SOLC VERSION }}.js"
+  src="https://binaries.soliditylang.org/bin/{{ SOLC_VERSION }}.js"
+  integrity="sha256-{{ BASE64_ENCODED_HASH_OF_SOLC_VERSION  }}"
+  crossorigin="anonymous"
 ></script>
 ```
 
-(Alternatively use `https://binaries.soliditylang.org/bin/soljson-latest.js` to get the latests version.)
+This will load `solc` into the global variable `window.Module`.
+Alternatively, use `https://binaries.soliditylang.org/bin/soljson-latest.js` to get the latest version.
```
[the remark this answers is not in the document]
matheusaaguiar: I thought about something like [latest version script](https://binaries.soliditylang.org/bin/soljson-latest.js), for example. Do you think this is appropriate?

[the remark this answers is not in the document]
r0qs: Oh right, my idea was to be more explicit about the location of the latest version since in the example above, it is shown as: `https://binaries.soliditylang.org/bin/{{ SOLC_VERSION }}.js`. And the link refers to an alternative to the example. But yes, we can omit the link if it makes it easier to understand.

Or maybe just: "Alternatively, replace `{{ SOLC_VERSION }}.js` by `soljson-latest.js` in the link above...`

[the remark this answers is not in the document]
matheusaaguiar: Understood. IMO, the latter suggestion is better, but just because I think that in general hiding the URLs behind an hyperlink is more elegant.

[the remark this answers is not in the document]
r0qs: But which latter suggestion? Yours or mine?
I think it should be clear that what we are referring to is the URL to be used in the script tag in the example. What about this one?

"Alternatively, use [soljson-latest.js](https://binaries.soliditylang.org/bin/soljson-latest.js) as `{{ SOLC_VERSION }}.js` in the code snippet above to load the latest version."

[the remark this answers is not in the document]
matheusaaguiar: I would suggest
```
Alternatively, use [soljson-latest.js](https://binaries.soliditylang.org/bin/soljson-latest.js) instead of {{ SOLC_VERSION }}.js in the code snippet above to load the latest version.
```

README.md
```diff
@@ -308,13 +321,37 @@ Add the version of `solc` you want to use into `index.html`:
...
-  src="https://binaries.soliditylang.org/bin/{{ SOLC VERSION }}.js"
+  src="https://binaries.soliditylang.org/bin/{{ SOLC_VERSION }}.js"
+  integrity="sha256-{{ BASE64_ENCODED_HASH_OF_SOLC_VERSION  }}"
+  crossorigin="anonymous"
 ></script>
 ```
 
-(Alternatively use `https://binaries.soliditylang.org/bin/soljson-latest.js` to get the latests version.)
+This will load `solc` into the global variable `window.Module`.
+Alternatively, use `soljson-latest.js` as `{{ SOLC_VERSION }}.js` in the code snippet above to load the latest version.
```
matheusaaguiar, line 331:
```suggestion
Alternatively, use `soljson-latest.js` instead of `{{ SOLC_VERSION }}.js` in the code snippet above to load the latest version.
```

README.md
```diff
@@ -204,6 +204,9 @@ The low-level API is as follows:
 
 For examples how to use them, please refer to the README of the above mentioned solc-js releases.
 
+**Note**: These low-level functions remain available for compatibility reasons.
+However, they were superseded by the `compile()` function and are no longer required in newer versions. The functions `compileSingle`, `compileMulti` and `compileCallback` are always `null` in the newer versions.
```
cameel, line 208: Please clarify that they're `null` when you use a newer binary (might also be good to check which version specifically; was that 0.5.0?) rather than newer solc-js. This distinction might not be clear to users.

For some context: it is recommended to always use latest release of solc-js and it's supposed to be able to handle handle all the older binaries, down to 0.1.x. Might be good to mention that in the README too if it's not mentioned already.

README.md
```diff
@@ -239,10 +242,20 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
     // An error was encountered, display and quit
   } else {
     // NOTE: Use `solcSnapshot` here with the same interface `solc` has
+    // For example:
+    const output = solcSnapshot.compile(/* ... */)
   }
 });
 ```
 
+The version **must** be in the long format string.
```
cameel, line 251:
```suggestion
The version **must** be in the long format.
```

README.md
```diff
@@ -239,10 +242,20 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
...
+Thus, if you would like to use version `v0.8.17` you need to provide it, including the committed version of the release.
```
cameel, line 252:
```suggestion
Thus, if you would like to use version `v0.8.17` you need to include the commit hash of the release.
```

README.md
```diff
@@ -239,10 +242,20 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
...
+You can extract the long version string for each version from the [public available release list](https://binaries.soliditylang.org/bin/list.json).
```
cameel, line 253: I wanted to say that we should rather point people towards the new `emscripten-wasm32/` and `emscripten-asmjs/` dirs but I realized that it would be inconsistent because we're still using and mentioning `bin/` everywhere. So I guess it's out of scope of this PR. FYI I created an issue for it: #671.

README.md
```diff
@@ -308,13 +321,37 @@ Add the version of `solc` you want to use into `index.html`:
...
+It is recommended that you check the integrity of the resource being fetched before using it in your application.
+For that, you can use the [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) feature.
+Adding SRI configuration to your HTML script tag ensures that the resource will only be loaded by the browser if the cryptographic hashes match.
+
+You can generate the SRI hash yourself based on the base64-encoded version of the sha256 hash of the release.
+For example, after downloading version `v0.8.16+commit.07a7930e`, run:
+```bash
+shasum -b -a 256 soljson-v0.8.16+commit.07a7930e.js | awk '{ print $1 }' | xxd -r -p | base64
+J7KCDvk4BaZcdreUWklDJYLTBv0XoomFcJpR5kA2d8I=
+```
```
cameel, line 342: This looks as if the hash was a part of the command to run.
````suggestion
```bash
shasum -b -a 256 soljson-v0.8.16+commit.07a7930e.js | awk '{ print $1 }' | xxd -r -p | base64
```
```
J7KCDvk4BaZcdreUWklDJYLTBv0XoomFcJpR5kA2d8I=
```
````

File: README.md [hunk 13]
```diff
@@ -308,13 +321,37 @@ Add the version of `solc` you want to use into `index.html`:
...
+Alternatively, use `soljson-latest.js` as `{{ SOLC_VERSION }}.js` in the code snippet above to load the latest version.
+
+It is recommended that you check the integrity of the resource being fetched before using it in your application.
+For that, you can use the [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) feature.
+Adding SRI configuration to your HTML script tag ensures that the resource will only be loaded by the browser if the cryptographic hashes match.
+
+You can generate the SRI hash yourself based on the base64-encoded version of the sha256 hash of the release.
+For example, after downloading version `v0.8.16+commit.07a7930e`, run:
+```bash
+shasum -b -a 256 soljson-v0.8.16+commit.07a7930e.js | awk '{ print $1 }' | xxd -r -p | base64
```
cameel, line 340: I think `xxd` may be a bit obscure (on my system it comes with gvim :)). Personally, I do use it, but for our users, who are likely JS developers perhaps it would be better to suggest a JS one-liner via `node -e` instead?

cameel, line 340: Any reason not to use `sha256sum` instead? Makes it a bit more compact and easier to type.
```suggestion
sha256sum --binary soljson-v0.8.16+commit.07a7930e.js | awk '{ print $1 }' | xxd -revert -plain | base64
```

cameel:
Needs rebase for the hardhat job to pass.

README.md
cameel: Honestly, for me the underscores make it look like an existing variable rather than a placeholder so I wanted to suggest removing them :P

README.md
r0qs: No, no particular reason. Changed.

r0qs: > on my system it comes with gvim
Really? Weird! haha

Ok. I think that the use of `xxd` is fine for the sake of an example. But I updated it to use node as you suggested.

README.md
r0qs: Yes, `0.5.0+commit.1d4f565a`.

README.md
```diff
@@ -204,6 +204,11 @@ The low-level API is as follows:
 
 For examples how to use them, please refer to the README of the above mentioned solc-js releases.
 
+**Note**: These low-level functions remain available for compatibility reasons.
+However, they were superseded by the `compile()` function and are no longer required.
+Starting in version `0.5.0+commit.1d4f565a`, the functions `compileSingle`, `compileMulti`, and `compileCallback` are always `null` when using newer binary versions.
```
matheusaaguiar, line 209:
```suggestion
Starting from version `0.5.0+commit.1d4f565a`, the functions `compileSingle`, `compileMulti`, and `compileCallback` are always `null` when using newer binary versions.
```
Also, I find the last part, `when using newer binary versions`, a little bit confusing because it doesn't specify which versions are `newer` and that those are `solc` binaries, not `solc-js` (right?).

README.md
```diff
@@ -204,6 +204,11 @@ The low-level API is as follows:
...
+It is recommended to use the latest release of solc-js, but it should also handle all the older binaries down to `0.1.x`.
```
matheusaaguiar, line 210:
```suggestion
It is recommended to use the latest release of solc-js, but it should also handle all the older solc binaries down to `0.1.x`.
```
This feels confusing to me too, who should handle older binaries? `solc` or `solc-js` binaries?

r0qs: The `solc-js` should handle old `solc` binaries.

README.md
r0qs: Right. Fixed.

matheusaaguiar (approved):
LGTM.

README.md
```diff
@@ -239,10 +244,20 @@ solc.loadRemoteVersion('latest', function(err, solcSnapshot) {
...
     // NOTE: Use `solcSnapshot` here with the same interface `solc` has
+    // For example:
+    const output = solcSnapshot.compile(/* ... */)
   }
 });
 ```
 
+The version **must** be in the long format.
+Thus, if you would like to use version `v0.8.17` you need to include the commit hash of the release.
+You can extract the long version string for each version from the [public available release list](https://binaries.soliditylang.org/bin/list.json).
```
cameel, line 255:
```suggestion
You can extract the long version string for each version from the [publicly available release list](https://binaries.soliditylang.org/bin/list.json).
```

cameel (approved):
LGTM. Please remember to squash review fixes into the original commits before merging.

PR #659, opened 2022-09-17
Add CI job to check if updateBinary downloads the correct release

r0qs:
This PR addresses the comments made here https://github.com/ethereum/solc-js/pull/648#issuecomment-1218209344 and is part of the solution for the issue https://github.com/ethereum/solc-js/issues/632.
It adds a CI job to test if the current version is equal to the downloaded release when using the `updateBinary` command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ workflows:
       - truffle-sample-project
       - cli-smoke-test
       - solidity-solcjs-ext-test
+      - download-release-solc-test
 
 version: 2.1
 
@@ -352,6 +353,16 @@ jobs:
       - run: git clone --depth 1 "https://github.com/ethereum/solidity" solidity/
       - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
       - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)" "$(realpath ../solc-js/)"
+
+  download-release-solc-test:
+    docker:
+      - image: cimg/node:16.15
+    steps:
+      - show-npm-version
+      - provision-and-package-solcjs
+      - run:
+          name: Check solc version with the released version
+          command: cd solc-js && scripts/update-and-check-binary.sh
 
   node-v10:
     <<: *node-base
--- a/scripts/update-and-check-binary.sh
+++ b/scripts/update-and-check-binary.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+BASE_URL="https://binaries.soliditylang.org/bin"
+REPO_ROOT="$(dirname "$0")/.."
+
+function fail() {
+	echo -e "$@" >&2
+	exit 1
+}
+
+function check_version() {
+	current_version=$(node ./dist/solc.js --version | sed -En 's/^(.*).Emscripten.*/\1/p')
+
+	# Retrieve the correspondent released version
+	short_version=$(echo "${current_version}" | sed -En 's/^([0-9.]+).*\+commit\.[0-9a-f]+.*$/\1/p')
+	release_version=$(curl -s "${BASE_URL}/list.json"  | jq ".releases | .[\"${short_version}\"]" | tr -d '"' | sed -En 's/^soljson-v(.*).js$/\1/p')
+
+	# check if current version exists as release
+	if [ "${current_version}" != "${release_version}" ]; then
+		fail "Failed: version mismatch:\n [current]: ${current_version}\n [release]: ${release_version}"
+	fi
+
+	current_sha=$(shasum -b -a 256 ./soljson.js | awk '{ print $1 }')
+	release_sha=$(curl -s "${BASE_URL}/list.json" | jq ".builds[] | select(.longVersion == \"${release_version}\") | .sha256" | tr -d '"' | sed 's/^0x//')
+
+	# check if sha matches
+	if [ "${current_sha}" != "${release_sha}" ]; then
+		fail "Failed: checksum mismatch:\n [current]: ${current_sha}\n [release]: ${release_sha}"
+	fi
+}
+
+(
+	cd "${REPO_ROOT}"
+
+	# Remove previous soljson.js binary if exists
+	[[ -f soljson.js ]] && rm -f soljson.js
+
+	# Update soljson.js binary
+	npm run updateBinary
+	npm run build
+
+	# Check if current binary works
+	echo "contract C {}" > C.sol
+	node ./dist/solc.js C.sol --bin
+	[[ ! -f C_sol_C.bin ]] && fail "Failed: downloaded binary is not working"
+	rm -f C.sol C_sol_C.bin
+
+	check_version
+)
--- a/scripts/update-and-check-binary.sh
+++ b/scripts/update-and-check-binary.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 BASE_URL="https://binaries.soliditylang.org/bin"
 REPO_ROOT="$(dirname "$0")/.."
 
@@ -39,11 +41,11 @@ function check_version() {
 	npm run updateBinary
 	npm run build
 
+	check_version
+
 	# Check if current binary works
 	echo "contract C {}" > C.sol
 	node ./dist/solc.js C.sol --bin
-	[[ ! -f C_sol_C.bin ]] && fail "Failed: downloaded binary is not working"
+	[[ ! -f C_sol_C.bin ]] && fail "Failed: downloaded binary may not be working properly"
 	rm -f C.sol C_sol_C.bin
-
-	check_version
 )


scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+	if [ "${current_version}" != "${release_version}" ]; then
+		fail "Failed: version mismatch:\n [current]: ${current_version}\n [release]: ${release_version}"
+	fi
+
+	current_sha=$(shasum -b -a 256 ./soljson.js | awk '{ print $1 }')
+	release_sha=$(curl -s "${BASE_URL}/list.json" | jq ".builds[] | select(.longVersion == \"${release_version}\") | .sha256" | tr -d '"' | sed 's/^0x//')
+
+	# check if sha matches
+	if [ "${current_sha}" != "${release_sha}" ]; then
+		fail "Failed: checksum mismatch:\n [current]: ${current_sha}\n [release]: ${release_sha}"
```
cameel, line 30: Single char indents are completely unreadable to me. You could just as well not indent it at all :)
```suggestion
		fail "Failed: checksum mismatch:\n    [current]: ${current_sha}\n    [release]: ${release_sha}"
```

cameel, line 30: I'd make errors stand out more:
```suggestion
		fail "ERROR: Checksum mismatch.\n [current]: ${current_sha}\n [release]: ${release_sha}"
```

And why not include that part directly in `fail()`?

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+	fi
+
+	# check if the current version is the latest release
+	latest_version=$(curl -s "${BASE_URL}/list.json"  | jq ".latestRelease" | tr -d '"')
+	if [ "${short_version}" != "${latest_version}" ]; then
```
cameel, line 35: I think it would be best to compare the full version, including the commit hash.

cameel (changes requested):

.circleci/config.yml
```diff
@@ -353,6 +354,16 @@ jobs:
       - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
       - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)" "$(realpath ../solc-js/)"
 
+  download-release-solc-test:
+    docker:
+      - image: cimg/node:16.15
```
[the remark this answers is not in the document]
r0qs: Probably yes, but I did not test though. I will do it.

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+
+function check_version() {
+	current_version=$(node ./dist/solc.js --version | sed -En 's/^(.*).Emscripten.*/\1/p')
+
+	# Retrieve the correspondent released version
+	short_version=$(echo "${current_version}" | sed -En 's/^([0-9.]+).*\+commit\.[0-9a-f]+.*$/\1/p')
+	release_version=$(curl -s "${BASE_URL}/list.json"  | jq ".releases | .[\"${short_version}\"]" | tr -d '"' | sed -En 's/^soljson-v(.*).js$/\1/p')
+
+	# check if current version exists as release
+	if [ "${current_version}" != "${release_version}" ]; then
```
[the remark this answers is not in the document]
r0qs: Yes indeed. I usually use with curly bracket because it is safer, in case for some mistake you forget to add it when the variable is surrounded by strings, like you described.

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+		fail "Failed: version mismatch:\n [current]: ${current_version}\n [release]: ${release_version}"
+	fi
+
+	current_sha=$(shasum -b -a 256 ./soljson.js | awk '{ print $1 }')
+	release_sha=$(curl -s "${BASE_URL}/list.json" | jq ".builds[] | select(.longVersion == \"${release_version}\") | .sha256" | tr -d '"' | sed 's/^0x//')
```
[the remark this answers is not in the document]
r0qs: Definitely, I will do it.

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+	echo -e "$@" >&2
+	exit 1
+}
+
+function check_version() {
+	current_version=$(node ./dist/solc.js --version | sed -En 's/^(.*).Emscripten.*/\1/p')
+
+	# Retrieve the correspondent released version
+	short_version=$(echo "${current_version}" | sed -En 's/^([0-9.]+).*\+commit\.[0-9a-f]+.*$/\1/p')
+	release_version=$(curl -s "${BASE_URL}/list.json"  | jq ".releases | .[\"${short_version}\"]" | tr -d '"' | sed -En 's/^soljson-v(.*).js$/\1/p')
```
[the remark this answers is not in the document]
r0qs: Oh! Nice, I didn't know about the `--raw-output`. Thanks.
Sadly I cannot apply the suggestion because it was marked by git as outdated due to the acceptance of the previous suggestion. But I will add it.

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+
+	# check if current version exists as release
+	if [ "${current_version}" != "${release_version}" ]; then
+		fail "Failed: version mismatch:\n [current]: ${current_version}\n [release]: ${release_version}"
+	fi
```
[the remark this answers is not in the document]
r0qs: No, I'm not ignoring it here. The `current_version` is the longest version format (which includes the commit hash) of the solc-js binary in `./dist/solc-js,` and the `release_version` is the one that exists in the `releases` field in the `list.json`. The `releases` field is a dictionary in which keys are short versions of releases and values are the binary names (which includes the long version strings), but I could have filtered it by value instead.

The idea was to check if the current version is a known release. But yeah, indeed, maybe just checking the hash is sufficient.

scripts/update-and-check-binary.sh
r0qs: Yes, I do that in the first check. The 3 checks are:
1. check if the currently installed version is a known release (it may not be the latest)
2. check if the sha of both matches
3. check if the current is also the latest

Indeed the first check seems useless, and I will just check the 2 and 3, using the commit hash in 3.

scripts/update-and-check-binary.sh
```diff
@@ -0,0 +1,57 @@
...
+	# check if sha matches
+	if [ "${current_sha}" != "${release_sha}" ]; then
+		fail "Failed: checksum mismatch:\n [current]: ${current_sha}\n [release]: ${release_sha}"
+	fi
+
+	# check if the current version is the latest release
+	latest_version=$(curl -s "${BASE_URL}/list.json"  | jq ".latestRelease" | tr -d '"')
+	if [ "${short_version}" != "${latest_version}" ]; then
+		fail "Failed: version is not the latest release:\n [current]: ${short_version}\n [latest]: ${latest_version}"
+	fi
```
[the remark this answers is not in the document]
r0qs: Not sure. The `updateBinary` script already checks this, but as we move it to the CI, it may also make sense to check here.

scripts/update-and-check-binary.sh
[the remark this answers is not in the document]
cameel: I think it's fine to use it if you prefer that. I only skip it in that one (but very common) case for less visual noise.

scripts/update-and-check-binary.sh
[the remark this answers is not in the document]
cameel: Don't worry about git suggestion mechanism :) I actually rarely use the "commit suggestion" option and simply add the change myself in my editor because that's more convenient for me. Most of the time we're going to squash the commits it creates anyway.

scripts/update-and-check-binary.sh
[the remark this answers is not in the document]
cameel: Ah, ok my bad. I must have looked at the short version. `^soljson-v(.*).js$` looks like it indeed should grab the commit hash too.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+REPO_ROOT="$(dirname "$0")/.."
+LIST_FILE=/tmp/list.json
+
+function fail() {
+    echo -e "ERROR: $@" >&2
+    exit 1
+}
+
+function check_release_version() {
+    current_version=$1
```
cameel, line 15:
```suggestion
    local current_version="$1"
```

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+
+    curl --silent --fail "$BASE_URL/list.json" -o $LIST_FILE
+    [[ ! -f $LIST_FILE ]] && fail "Download of release list failed:\n    [url]: $BASE_URL/list.json"
+
+    # Retrieve the latest released version
+    latest_version=$(cat $LIST_FILE | jq --raw-output ".latestRelease")
```
cameel, line 21:
```suggestion
    latest_version=$(cat "$LIST_FILE" | jq --raw-output ".latestRelease")
```
It's unquoted also in a few other places. `$current_version` as well.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+function fail() {
+    echo -e "ERROR: $@" >&2
+    exit 1
+}
+
+function check_release_version() {
+    current_version=$1
+
+    curl --silent --fail "$BASE_URL/list.json" -o $LIST_FILE
+    [[ ! -f $LIST_FILE ]] && fail "Download of release list failed:\n    [url]: $BASE_URL/list.json"
```
cameel, line 18: I see you removed the curly braces everywhere :) That's not really what I meant. I only said they are kinda superfluous if the variable stands on its own. I do agree that it's a good practice to have them when it's a part of a string though:
```suggestion
    [[ ! -f $LIST_FILE ]] && fail "Download of release list failed:\n    [url]: ${BASE_URL}/list.json"
```

cameel: By the way, maybe we should add something about shell style in [`CODING_STYLE.md`](https://github.com/ethereum/solidity/blob/develop/CODING_STYLE.md)?

Unless we can configure shellcheck to catch these automatically. We should at least have shellcheck in solc-js too.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+BASE_URL="https://binaries.soliditylang.org/bin"
+REPO_ROOT="$(dirname "$0")/.."
+LIST_FILE=/tmp/list.json
```
cameel, line 7: Safer to ask the OS to generate the file. Temp dir might not be `/tmp` or `list.json` might already exist there.
```suggestion
LIST_FILE=$(mktemp -t solc-bin-list-XXXXXX.json)
```

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+    # Check if current version is the latest release
+    if [ $current_version != "$release_version" ]; then
+        fail "Version is not the latest release:\n    [current]: $current_version\n    [latest]: $latest_version"
+    fi
+
+    current_sha=$(shasum --binary --algorithm 256 ./soljson.js | awk '{ print $1 }')
+    release_sha=$(cat $LIST_FILE | jq --raw-output ".builds[] | select(.longVersion == \"$release_version\") | .sha256" | sed 's/^0x//')
+
+    # Check if sha matches
+    if [ "$current_sha" != "$release_sha" ]; then
```
cameel, line 33:
```suggestion
    if [[ $current_version != "$release_version" ]]; then
        fail "Version is not the latest release:\n    [current]: $current_version\n    [latest]: $latest_version"
    fi

    current_sha=$(shasum --binary --algorithm 256 ./soljson.js | awk '{ print $1 }')
    release_sha=$(cat $LIST_FILE | jq --raw-output ".builds[] | select(.longVersion == \"$release_version\") | .sha256" | sed 's/^0x//')

    # Check if sha matches
    if [[ $current_sha != "$release_sha" ]]; then
```

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+
+function check_release_version() {
+    current_version=$1
+
+    curl --silent --fail "$BASE_URL/list.json" -o $LIST_FILE
+    [[ ! -f $LIST_FILE ]] && fail "Download of release list failed:\n    [url]: $BASE_URL/list.json"
+
+    # Retrieve the latest released version
+    latest_version=$(cat $LIST_FILE | jq --raw-output ".latestRelease")
+    release_version=$(cat $LIST_FILE | jq --raw-output ".releases | .[\"$latest_version\"]" | sed --regexp-extended --quiet 's/^soljson-v(.*).js$/\1/p')
```
cameel, line 22: They're both the latest version, aren't they? For me this naming would be clearer:
```suggestion
    latest_version_short=$(cat $LIST_FILE | jq --raw-output ".latestRelease")
    latest_version_long=$(cat $LIST_FILE | jq --raw-output ".releases | .[\"$latest_version_short\"]" | sed --regexp-extended --quiet 's/^soljson-v(.*).js$/\1/p')
```

cameel, line 22: I think it would be more correct to get the version from the `longVersion` field of the `builds` list instead of parsing it out of the path. Technically, the path could be arbitrary and we just happen to put it in the file name but we don't have to.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+
+    current_version=$(node ./dist/solc.js --version | sed --regexp-extended --quiet 's/^(.*).Emscripten.*/\1/p')
+
+    # Verify if current version matches the package version.
+    # It already exits with an error if the version mismatch
+    node ./dist/verifyVersion.js
+
+    # Verify if current version is the latest release
+    check_release_version $current_version
+    if [ $? -eq 0 ]; then
```
cameel, line 49:
```suggestion
    if check_release_version "$current_version"; then
```
But due to `set -e` a non-zero code will interrupt the script anyway so we could just as well skip the `if`.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+        echo "solc-js version $current_version is the latest release"
+    fi
+
+    # Cleanup temp files
+    [[ -f $LIST_FILE ]] && rm -f $LIST_FILE
```
cameel, line 54: Why check for existence if you're using `--force` anyway?
```suggestion
    rm -f "$LIST_FILE"
```

.circleci/config.yml
```diff
@@ -353,6 +354,20 @@ jobs:
       - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
       - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)" "$(realpath ../solc-js/)"
 
+  download-release-solc-test:
+    docker:
+      - image: cimg/node:current
+    steps:
+      - show-npm-version
+      - provision-and-package-solcjs
```
cameel, line 362: Do we need this step? You seem to be using the repo directly, rather than the package this installs.

scripts/check-binary.sh
```diff
@@ -0,0 +1,55 @@
...
+    current_version=$(node ./dist/solc.js --version | sed --regexp-extended --quiet 's/^(.*).Emscripten.*/\1/p')
+
+    # Verify if current version matches the package version.
+    # It already exits with an error if the version mismatch
+    node ./dist/verifyVersion.js
+
+    # Verify if current version is the latest release
+    check_release_version $current_version
+    if [ $? -eq 0 ]; then
+        echo "solc-js version $current_version is the latest release"
```
cameel, line 50: How about this?
```suggestion
        echo "The currently installed soljson.js binary (${current_version}) matches the latest release available in solc-bin."
```

.circleci/config.yml
```diff
@@ -353,6 +354,20 @@ jobs:
       - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
       - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)" "$(realpath ../solc-js/)"
 
+  download-release-solc-test:
```
cameel, line 357:
```suggestion
  update-binary-test:
```

.circleci/config.yml
```diff
@@ -353,6 +354,20 @@ jobs:
...
+    steps:
+      - show-npm-version
+      - provision-and-package-solcjs
+      - run:
+          name: Check solc version with the released version
+          command: |
+            cd solc-js
+            npm run updateBinary
+            npm run build
+            scripts/check-binary.sh
```
cameel, line 369: "check" is not very specific. Can we give it a better name?
```suggestion
          name: Verify that `npm run updateBinary` downloads the latest release
          command: |
            cd solc-js
            npm run updateBinary
            npm run build
            scripts/is-binary-up-to-date.sh
```

cameel (commented):
Some more minor tweaks but the overall direction of the PR is good.

scripts/check-binary.sh
r0qs: Oh, I did not know about `mktemp`, nice!

scripts/check-binary.sh
r0qs: Indeed.

scripts/check-binary.sh
r0qs: Yes, it would be great indeed. Are there any of the default rules usually ignored by you? Or should we stick with their general recommendations?
We can add shellcheck as a pre-commit hook for `.sh` files (https://github.com/koalaman/shellcheck/blob/master/README.md#pre-commit).

scripts/check-binary.sh
r0qs: The problem is that the `latestRelease` is a short version, and I don't know beforehand what is the correspondent long version of the release to be able to disambiguate in the builds list. So I needed to extract that information from the `releases` field and filter the correct entry in the `builds`.

r0qs: But yes, I guess it is better to invert the logic, try to find the current release (of which I know the long version) in the builds list, as you said, and then see if its short version is the latest release.

.circleci/config.yml
r0qs: No, it is not. I used it because of the provision part. I thought about installing it as a dependency but changed my mind and did not update the script.

scripts/check-binary.sh
cameel: Their general recommendations are very good. I can actually only think of two I'd disable globally:
- [SC2016: Expressions don't expand in single quotes, use double quotes for that](https://www.shellcheck.net/wiki/SC2016): This is useless because it has tons of false-positives. It basically does not let you use single quotes inside double quotes ever, even in valid cases. And there's no "fix", you have to ignore the warning.
- [SC2002: Useless cat. Consider `cmd < file | ..` or `cmd file | ..` instead](https://www.shellcheck.net/wiki/SC2002): I know it's theoretically less efficient but in practice this does not matter and I find the `cat file | cmd1 | cmd2` form much more logical and readable. I prefer it over something like `cmd1 < file | cmd2` and `< file cmd1 | cmd2` just looks weird.

scripts/is-binary-up-to-date.sh
```diff
@@ -0,0 +1,54 @@
...
+
+    curl --silent --fail "${BASE_URL}/list.json" -o "$LIST_FILE"
+    [[ ! -f $LIST_FILE ]] && fail "Download of release list failed:\n    [url]: ${BASE_URL}/list.json"
+
+    # Retrieve the latest released version
+    latest_version_short=$(jq --raw-output ".latestRelease" "$LIST_FILE")
+    latest_version_long=$(jq --raw-output ".releases | .[\"${latest_version_short}\"]" "$LIST_FILE" | sed --regexp-extended --quiet 's/^soljson-v(.*).js$/\1/p')
+
+    # Check if current version is the latest release
+    if [[ "$current_version" != "$latest_version_long" ]]; then
```
cameel, line 25: As a general rule, inside `[[ ]]` you only need the quotes on the right side. And that's only in cases where the stuff on the right side could be mistakenly interpreted as a regex (quoting it disables that).
```suggestion
    if [[ $current_version != "$latest_version_long" ]]; then
```

cameel (approved):
There's still https://github.com/ethereum/solc-js/pull/659#discussion_r978673390 left to resolve but this is not a major issue so I'm already approving. Fine to merge whether you decide to go with my suggestion there or not.

scripts/check-binary.sh
cameel: Good point about it being ambiguous. I did not notice that.

But in that case you can still use the path you get from `.releases` to find the right entry in `.builds` and avoid parsing that path with a regex. Paths are supposed to be unique and it will be more robust.

cameel:
Before merging please squash the commits a bit. Does not have to be just one but in this case this seems like the best option.It all looks like review fixes and we usually squash those into the original commits they change.

scripts/check-binary.sh
r0qs: The point there was to compare only the long versions between `current_version` and `latest_version_long` ignoring some path elements like `soljson-v` or `.js`. But I applied the changes you suggested and used the `path` to avoid the extra regex.

PR #662, opened 2022-09-24
Add CI job to check coding style

r0qs:
This PR adds a CI job to perform coding style checks as suggested here: https://github.com/ethereum/solc-js/pull/659#discussion_r978652925

It performs the `eslint` checks for `.js` and `.ts` files using the npm lint script and checks the `.sh` files using the [shellcheck circleCI orb](https://circleci.com/developer/orbs/orb/circleci/shellcheck).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ workflows:
   version: 2.1
   node-multi-build:
     jobs:
+      - check-coding-style
       - node-v10
       - node-v12
       - node-v14
@@ -16,6 +17,9 @@ workflows:
       - solidity-solcjs-ext-test
 
 version: 2.1
+
+orbs:
+  shellcheck: circleci/shellcheck@volatile
 
 commands:
   show-npm-version:
@@ -181,6 +185,24 @@ jobs:
             - run:
                 name: coveralls
                 command: npm run coveralls
+
+  check-coding-style:
+    docker:
+      - image: cimg/node:current
+    steps:
+      - show-npm-version
+      - checkout
+      - shellcheck/install
+      - install-dependencies:
+          cache-id: solc-js
+      - run:
+          name: Check for javascript/typescript coding style
+          command: npm run lint
+      - shellcheck/check:
+          ignore-dirs: |
+            ./.git
+            ./node_modules
+            ./dist
 
   hardhat-core-default-solc:
     # Runs out of memory on 'medium'.


.circleci/config.yml
```diff
@@ -182,6 +186,24 @@ jobs:
...
+      - install-dependencies:
+          cache-id: solc-js
+      - run:
+          name: Check for javascript/typescript coding style
+          command: npm run lint
+      - shellcheck/check:
+          ignore-dirs: |
+            ./.git
+            ./node_modules
+            ./dist
```
cameel, line 205: Isn't `./` superfluous? On the other hand I'd add `/` at the end to make it clear we expect these to be dirs.
```suggestion
            .git/
            node_modules/
            dist/
```

cameel (approved):
I have some minor suggestions but looks good overall.

.circleci/config.yml
```diff
@@ -17,6 +18,9 @@ workflows:
 
 version: 2.1
 
+orbs:
+  shellcheck: circleci/shellcheck@3.1.1
```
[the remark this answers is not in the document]
r0qs: Definitely! If there are many repeated tasks there that could also be reused by other repos in the organization, it would be beneficial to split them in orbs.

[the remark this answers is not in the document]
r0qs: Yes. I can do it.
Hum...it is not recommended to use the [latest version](https://circleci.com/docs/orbs-faq#how-to-use-the-latest-version-of-an-orb) since it can introduce unexpected breaking changes, but you can set it as `@volatile` to always use the latest if needed. But we can set only the major version as well, so circleCI will always use the latest compatible version up to that major version.

.circleci/config.yml
r0qs: Yes. I was wondering if we indeed need to exclude those directories because they will probably not contain any shell script. But to avoid the orb to even parsing those directories, I preferred to explicitly them.

.circleci/config.yml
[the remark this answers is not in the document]
cameel: I'd just use `volatile` in case of shellcheck specifically since I don't expect much heavy breakage here but in other cases like node.js major version would be fine too.

Breaking changes are not such a big issue for tools we use in CI. We generally prefer to stay on the latest versions and just fix some minor problems they might introduce from time to time. And if the breakage is bigger, we can always hard-code it then. It's a trade-off but I think it's better than using an ancient version and then feeling the pain of upgrade all at once :) With versions hard-coded everywhere it's also hard to keep them up to date manually.

.circleci/config.yml
cameel: You could also create a separate job for shellcheck. Then you would not have to install the JS dependencies in it. Also, the shellcheck job would probably complete faster.

By the way, you probably don't need to ignore `dist/` because you're not running `build`.

.circleci/config.yml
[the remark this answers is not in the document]
cameel: There is some common stuff. For example the [gitter notification command](https://github.com/ethereum/solidity/blob/v0.8.17/.circleci/config.yml#L34-L93) may be a good candidate for an orb (while at it, we could also make them work on Windows). The [`t_ems_ext`](https://github.com/ethereum/solidity/blob/v0.8.17/.circleci/config.yml#L1210-L1271) is used by external tests and [their definitions](https://github.com/ethereum/solidity/blob/v0.8.17/.circleci/config.yml#L538-L650) are pretty repetitive too. And there are some common [steps/templates](https://github.com/ethereum/solidity/blob/v0.8.17/.circleci/config.yml#L118-L285). Also generally splitting the whole file into smaller bits by grouping similar jobs (build/test jobs, bytecode compare, external tests, etc.) would be great if possible.

We have a whole issue about that (https://github.com/ethereum/solidity/issues/11846) but there are limits to what we can do and so far I did not consider creating orbs.

.circleci/config.yml
```diff
@@ -182,6 +186,24 @@ jobs:
...
+      - install-dependencies:
+          cache-id: solc-js
+      - run:
+          name: Check for javascript/typescript coding style
+          command: npm run lint
+      - shellcheck/check:
+          ignore-dirs: |
+            .git/
+            node_modules/
+            dist/
```
cameel, line 205: Maybe shellcheck does need the `./` bit after all? I see in the CI output that it did not skip files inside node_modules:

```
./node_modules/tape/example/static/build.sh
./node_modules/foreground-child/changelog.sh
No ShellCheck Errors Found
```
Unless it just prints them without checking?

cameel: By the way, the orb is nice. I see that it for example automatically adds `shellcheck.log` as an artifact (if non-empty). That's handy.

.circleci/config.yml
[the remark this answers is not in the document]
r0qs: Nice! Thanks, I can take a look on that :)
I can start with the gitter notification then, since there is already an issue for the failing solc-bin nightly builds.

.circleci/config.yml
r0qs: oh no! Indeed it is not skipping those directories. There is an example of the config in the official repo: https://github.com/CircleCI-Public/shellcheck-orb/blob/73e2427de6b0a5d0c27527b608ee62880e3dbfb0/.circleci/config.yml#L22 and the `./` is used. I will put it back.

r0qs: Hum...still it shows. So yeah, it seems that it parses anyway but probably don't check as you mentioned.

cameel: Can you try it out locally on a file where you intentionally make a mistake to make sure?

r0qs: Yes. It seems that it indeed considers the "ignored" folders when they terminate with `/` (e.g. `node_modules/`). If I remove the slash in the end, it works.

With `node_modules/` or `./node_modules/`:
```
./scripts/wrong.sh
./node_modules/foreground-child/changelog.sh
./node_modules/tape/example/static/build.sh
ShellCheck Errors Found

In ./scripts/wrong.sh line 5:
if [ $1 -eq "1"]; then
^-- SC1009: The mentioned syntax error was in this if expression.
   ^-- SC1073: Couldn't parse this test expression. Fix to allow more checks.
                ^-- SC1020: You need a space before the ].
                ^-- SC1072: Missing space before ]. Fix any mentioned problems and try again.
```

Without the slash in the end `./node_modules`:
```
./scripts/wrong.sh
ShellCheck Errors Found

In ./scripts/wrong.sh line 5:
if [ $1 -eq "1"]; then
^-- SC1009: The mentioned syntax error was in this if expression.
   ^-- SC1073: Couldn't parse this test expression. Fix to allow more checks.
                ^-- SC1020: You need a space before the ].
                ^-- SC1072: Missing space before ]. Fix any mentioned problems and try again
```

In conclusion, we need the current directory prefix, and we should not add the end slash.

cameel: ok. Let's revert to the way it was originally.

Too bad that it's that picky. In the shell these would be perfectly fine.

r0qs: Yes. I reverted it already. I opened an issue in the orb repo anyway, asking about this behavior.

cameel:
So I guess we're done here? Looks ready to merge to me.

cameel:
Oh, please squash the commits a bit though. Looks like 3 of the 4 commits are just review tweaks to the first one so they should be squashed into it.

Issue #581, opened 2022-01-07
"Browser Usage" in Readme probably wrong due to synchronous webassembly compilation


ekpyron:
We specifically compile ``soljson.js`` with ``WASM_ASYNC_COMPILATION=0`` - there was a reason for this, but I don't remember right now.
However, I think at least chromium-based browsers won't allow synchronous compilation from the main thread. This effectively means that loading the compiler *has* to be done from web workers right now and the "Browser Usage" section in ``README.md`` is outdated.

So we need to check if we can switch to asynchronous compilation in solidity, investigate if and what kind of problems synchronous compilation entails and adjust the section in the readme accordingly.

jhubbardsf:
Any updates on this?

PR #686, opened 2023-03-13
Add web worker example in README

r0qs:
This PR adds a note in the README about the need for the use of web workers when using `solc` in web browsers as suggested by https://github.com/ethereum/solc-js/issues/581 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,64 +318,47 @@ var output = translate.prettyPrintLegacyAssemblyJSON(assemblyJSON, sourceCode)
 
 ## Browser Usage
 
-Add the version of `solc` you want to use into `index.html`:
+Compilation is generally a long-running and resource intensive task that cannot reasonably be performed in the main thread of the browser.
+Some browsers even dissallow synchronous compilation on the main thread if the module is larger than 4KB.
+Thus, the only supported way to use `solc` in a web browser is through a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers).
 
+### Loading solc with web workers
+
+Web Workers allow you to run javascript in the background in the browser, letting the browser's main thread free to do whatever it needs to do.
+Please, see the minimal example of how to use `solc` with web workers below or check out this [repository](https://github.com/r0qs/solcjs-webworker-example) for a full demo.
+
+* index.html
 ```html
-<script
-  type="text/javascript"
-  src="https://binaries.soliditylang.org/bin/{{ SOLC_VERSION }}.js"
-  integrity="sha256-{{ BASE64_ENCODED_HASH_OF_SOLC_VERSION  }}"
-  crossorigin="anonymous"
-></script>
+<!DOCTYPE html>
+<html>
+
+<head>
+	<meta charset="utf-8" />
+</head>
+
+<body>
+	<script>
+		var worker = new Worker('./dist/bundle.js');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -318,6 +318,14 @@ var output = translate.prettyPrintLegacyAssemblyJSON(assemblyJSON, sourceCode)
 
 ## Browser Usage
 
+**NOTE**: We currently compile the wasm module in `soljson.js` synchronously (see [here](https://github.com/ethereum/solc-js/issues/581) for more information), which may not work in Chrome-based
+browsers since it limits the size of modules that can be compiled/fetched synchronously to avoid blocking the main thread.
+Thus, if you get an error like: `WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB`,
+when trying to use the method described [below](#directly-loading-the-module-does-not-work-on-chrome-based-browsers),
+please consider trying to use `solc` through a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) as described [here](#loading-solc-with-web-workers) instead.
+
+### Directly loading the module (does not work on chrome-based browsers)
```
ekpyron, line 327: Should we just remove this section entirely and suggest to *always* use web workers generally? People will rarely want to do anything that only works on a fraction of browsers anyways, right?

Fine with keeping it as well, though.

r0qs: Right. Indeed, maybe it is better to just recommend to always use web workers as you said. This can also avoid some confusions. I will update the readme.

README.md
```diff
@@ -318,64 +318,47 @@ var output = translate.prettyPrintLegacyAssemblyJSON(assemblyJSON, sourceCode)
 
 ## Browser Usage
 
-Add the version of `solc` you want to use into `index.html`:
+We currently compile the wasm module in `soljson.js` synchronously (see [here](https://github.com/ethereum/solc-js/issues/581) for more information), which may not work in Chrome-based
+browsers since it limits the size of modules that can be compiled/fetched synchronously to avoid blocking the main thread.
+Thus, the recommended way to use `solc` in a web browser is through a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers).
```
ekpyron, line 323: I think we can be even more aggressive with this.
I.e. something along the lines of ``Compilation is generally a long-running and resource intensive task that cannot reasonably be performed in the main thread. The only supported way to use `solc` in a web browser is through a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers).``

I actually think browsers would even complain about asynchronously running the compiler due to it running too long and the browser suspecting it hangs - but I'm not 100% sure that's true :-). Web workers are really the proper way to do this in general, aren't they?

r0qs: Firefox does not complain about it, at least not on the simple tests that I did. But chromium immediately complains emitting an warning that `WebAssembly.Compile` is disallowed on the main thread if the buffer size is larger than 4KB.

In general I do agree that web workers are the proper way to do this, although some web browsers will still allow you to compile web assembly in the browser's main thread.

ekpyron: I know - but what's the use case for anything that works in *some*, but not all browsers :-)? Resp. "supported use" would have to mean something that works universally. I.e. as long as something not works in *all* browsers, I don't think it makes sense to suggest it in any way.

ekpyron (approved):

PR #703, opened 2023-07-18
Set version to 0.8.21

nikola-matic:


r0qs (approved):
Tested on Remix and works as expected :)

PR #706, opened 2023-09-04
Create node.js.yml

Ishmaello:


r0qs:
Any specific reason to suggest this? Also, the PR is not doing anything relevant and even uses wrong npm scripts. We already use Circle CI for the builds.

Ishmaello:
Then I delete,I think all this havork is because I'm looking for a solidity(sol) file to verify my contact

PR #716, opened 2023-11-10
Bump node versions in CI

cameel:
This bumps all minor versions to latest, and also adds a new run on `node:current` (which is node 21).

It also bumps node version in Hardhat jobs that were not on `current` only because at the time of node 17 Hardhat did not work with it.

Truffle does not work with node 21, but is fine with node 20, which is the current `lts`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,8 @@ workflows:
       - node-v14
       - node-v16
       - node-v18
-      - node-v20:
+      - node-v20
+      - node-current:
           run_coveralls: true
       - hardhat-core-default-solc
       - hardhat-core-latest-solc
@@ -397,3 +398,7 @@ jobs:
     <<: *node-base
     docker:
       - image: cimg/node:20.5
+  node-current:
+    <<: *node-base
+    docker:
+      - image: cimg/node:current
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,7 @@ workflows:
       - node-v10
       - node-v12
       - node-v14
-      - node-v16:
-          run_coveralls: true
+      - node-v16
       - node-v18
       - node-v20:
           run_coveralls: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,9 +173,6 @@ jobs:
       - install-dependencies:
           cache-id: solc-js
       - run:
-          name: install-npm
-          command: npm install
-      - run:
           name: updateBinary
           command: npm run updateBinary
       - run:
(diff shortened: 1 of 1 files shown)


cameel:
BTW, I see we still have a branch called `bump-truffle-sample-project-to-node-18`, that was apparently created during 0.8.22 release. @nikola-matic I assume it's safe to delete it now?

r0qs (commented):
Nice, just one suggestion. Maybe we could have a docker image parameter, like we have in the solidity repo, but with the default nodejs version set to `cimg/node:current` so we would only set specific versions where needed and reduce code repetition, since most of the places uses `node:current` already.

```yaml
parameters:
  default-nodejs-docker-image:
    type: string
    default: "cimg/node:current"
...
  hardhat-sample-project:
    docker:
      - image: << pipeline.parameters.default-nodejs-docker-image >>
```

Or even something like:

```yaml
nodejs-defaults: &nodejs-defaults
  docker:
    - image: cimg/node:current
...
hardhat-sample-project:
  <<: *nodejs-defaults
  steps:
  ...
```

.circleci/config.yml
```diff
@@ -172,9 +172,6 @@ jobs:
       - checkout
       - install-dependencies:
           cache-id: solc-js
-      - run:
-          name: install-npm
-          command: npm install
```
r0qs, line 177: Was this removed in favor of `install-dependencies`?

cameel:
> Maybe we could have a docker image parameter, like we have in the solidity repo

Not sure it's much of an improvement. We use specific versions only once now and the rest is `cimg/node:current`, which is already almost like a variable.

In the more repo it makes more sense because versions are long due to the embedded hash.

.circleci/config.yml
cameel: It's superfluous. `install-dependencies` already either installs required packages or restores them from cache.

I'm not sure why I included this originally, and I suspect this is just a leftover from before I refactored dependency install into a command.

The command that would actually make *some* sense here would be `npm update` which would update to latest version where available (the cached ones get updated only when we bump version or update `package.json` due to some other reason) but this makes CI less predictable and defeats the purpose of caching them to some extent.

r0qs:
> > Maybe we could have a docker image parameter, like we have in the solidity repo
> 
> Not sure it's much of an improvement. We use specific versions only once now and the rest is `cimg/node:current`, which is already almost like a variable.
> 
> In the more repo it makes more sense because versions are long due to the embedded hash.

Fair enough. Looks good to me then.

r0qs (approved):

Issue #715, opened 2023-11-09
0.8.23 npm release won't install on Windows due to a stray dependency


Janther:
The last release has a dependency `"n": "^9.2.0"` and is blocking the installation of this package on windows machines since it [does not support](https://github.com/tj/n#supported-platforms) that.

Oddly enough there is no mention of this package in the repo.

<img width="804" alt="Screen Shot 2023-11-09 at 5 08 40 PM" src="https://github.com/ethereum/solc-js/assets/1022054/2a10fff5-bb42-4b0a-a662-8a1042de5c77">

<img width="265" alt="Screen Shot 2023-11-09 at 5 09 00 PM" src="https://github.com/ethereum/solc-js/assets/1022054/2a3beb14-ddd0-481b-87f8-12179aefc8b9">

Janther:
if I look at the [release on GitHub](https://github.com/ethereum/solc-js/releases/tag/v0.8.23), it doesn't have that dependency either but it does when I look at the [npm package](https://www.npmjs.com/package/solc)

Janther:
I reverted the bump on `solc@0.8.23` so we can safely publish a version that runs on all machines

cameel:
Yes, that dependency should not be there. It must have been introduced when the release package was being built.

Unfortunately npm does not allow reuploading packages so fixing this will require a new release. For now downgrading to 0.8.22 sounds like a good workaround.

matheusaaguiar:
Hi @Janther , unfortunately this dependency was accidentally inserted and although unrelated and harmless, it doesn't work in Windows and thus create the problem you reported.
We have released [`0.8.23-fixed`](https://www.npmjs.com/package/solc/v/0.8.23-fixed) to address that until we the next release.

PR #717, opened 2023-11-10
Building release package in CI

cameel:
The long-term solution for problems like #715.

This PR extracts the building of the .tgz package into a separate CI job that's now a dependency of other jobs that use that package. It also adjusts those jobs to rely only on the package and not reach into the checked out repo (that used to be available them).

The package is now a CI artifact and we can use it in our release process instead of building it locally, which is error prone.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,8 @@
+defaults:
+  - requires_package: &requires_package
+      requires:
+        - build-package
+
 workflows:
   version: 2.1
   node-multi-build:
@@ -11,11 +16,12 @@ workflows:
       - node-v20
       - node-current:
           run_coveralls: true
-      - hardhat-core-default-solc
-      - hardhat-core-latest-solc
-      - hardhat-sample-project
-      - truffle-sample-project
-      - cli-smoke-test
+      - build-package
+      - hardhat-core-default-solc: *requires_package
+      - hardhat-core-latest-solc: *requires_package
+      - hardhat-sample-project: *requires_package
+      - truffle-sample-project: *requires_package
+      - cli-smoke-test: *requires_package
       - solidity-solcjs-ext-test
 
 version: 2.1
@@ -104,7 +110,7 @@ commands:
         default: .
       tarball-path:
         type: string
-        default: solc-js.tgz
+        default: workspace/solc-js.tgz
       package-manager:
         type: enum
         enum: ["npm", "yarn", "pnpm"]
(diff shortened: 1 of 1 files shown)


cameel:
When we merge this, we should make the `build-package` job mandatory.

.circleci/config.yml
```diff
@@ -240,24 +267,27 @@ jobs:
...
-            HARDHAT_TESTS_SOLC_PATH="${PWD}/solc-js/soljson.js"
-            HARDHAT_TESTS_SOLC_VERSION=$(jq --raw-output .version solc-js/package.json)
+            cd hardhat/
+            HARDHAT_TESTS_SOLC_PATH="${PWD}/node_modules/solc/soljson.js"
+            HARDHAT_TESTS_SOLC_VERSION=$(jq --raw-output .version node_modules/solc/package.json)
             export HARDHAT_TESTS_SOLC_PATH HARDHAT_TESTS_SOLC_VERSION
 
-            cd hardhat/packages/hardhat-core
+            cd packages/hardhat-core/
             yarn test
```
r0qs, line 282: If we update the `cimg/node` version to 18+ we can use pnpm here, which is installed by default in that image but not for the `16.15`. Because I suspect this will start to fail with the latest hardhat update.

```suggestion
            pnpm test
```

.circleci/config.yml
```diff
@@ -240,24 +267,27 @@ jobs:
       - image: cimg/node:16.15
```
r0qs, line 267: We can update this to be the same used in the solidity repo.
```suggestion
      - image: cimg/node:18.16
```

.circleci/config.yml
r0qs: Then we can change `provision-hardhat-with-packaged-solcjs` to use `pnpm` as `package-manager`. Adding `pnpm` as a `package-manager` option to the `inject-solc-js-tarball` command. What do you think?

r0qs (commented):
Nice. I have just some comments regarding the latest hardhat changes that broke our CI in the solidity repo. I can push the suggested refactors as fixup if you think they make sense :)

r0qs:
Oh, just remembered that you have another PR bumping the node versions (https://github.com/ethereum/solc-js/pull/716), so yeah, we can merge that first and then update the command to also have the `pnpm` as package-manager.

.circleci/config.yml
cameel: I merged #716 so versions are up to date now.

cameel:
Well, the transition to pnpm will probably be more tricky here, because we're doing more unusual stuff (i.e. injecting the whole local solc-js, not just the binary).

cameel:
So looks like this needs more work to pass CI. I'm busy with other PRs now so I'll go back to this later.

cameel:
Actually, the pnpm thing is also independent of this PR so also feel free to just take the top commit and try to fix it on `master` if you want to try.

r0qs:
> Actually, the pnpm thing is also independent of this PR so also feel free to just take the top commit and try to fix it on `master` if you want to try.

I created a PR here https://github.com/ethereum/solc-js/pull/719. But I needed to do some workarounds to make it work with pnpm, not really sure if it is the best way.

r0qs (approved):
I'm approving although the CI will not pass until we merge https://github.com/ethereum/solc-js/pull/719

Issue #570, opened 2021-12-14
package.json should use fixed versions.


MicahZoltu:
https://github.com/ethereum/solc-js/blob/ee8632967bf1a51602753e113897466e6874be7e/package.json#L47-L57

To protect users against supply chain attacks, package.json should refer to explicit versions and version updates should be done intentionally rather than "when CI happens to build" or "when user happens to install the dependency".

It is worth noting that this does *not* protect users against transitive dependencies of this library, unless each of the dependencies *also* uses explicit versions in `package.json`.  To more thoroughly protect users against supply chain attacks, only dependencies that also use explicit dependency versions should be used and the same for their dependencies etc.

---

Separately, I think it is worth discussing the use of `npm shrinkwrap`.  This would make it so the developers of this library can do a one-time audit of their dependency tree and then commit and publish that so downstream users don't have to each do a full audit of `solcjs`'s transitive dependencies every time they update or install it.

chriseth:
I was not aware that the dependencies grew that much. Maybe it would be good to also reduce the non-dev-dependencies to a minimal set. I don't think anybody ever does a full review of all the libraries.

MicahZoltu:
https://github.com/ethereum/solc-js/issues/361 is tracking the reduction of dependencies.  It is down a long way from when that issue was opened, but I suspect more progress could be made if someone dug into it a bit.

I do wonder if some of those could be moved to `devDependencies`?  For example, semver maybe is used by the build pipeline but not runtime?

cameel:
Thanks for opening the discussion. 

The biggest problem I see here is that if we fix these versions, we force any downstream project to use the same exact versions and prevent them from updating. IMO the library should be as agnostic to that as possible. I.e. if we know it won't work correctly with a specific version of a specific dependency, we should adjust the range in `package.json` to exclude that version but other than that we should adapt to what downstream wants to use.

I think that many downstream projects (e.g. Truffle or Hardhat) are much more proactive about keeping their dependencies up to date than we are and it might be be more of a nuisance than help for them. I think that hard-coding of specific versions would only work if we could commit to keeping them always up to date and we currently and I don't think we have bandwidth for that.

MicahZoltu:
A downstream user can override versions if they want.  For everything cryptocurrency, I think there is very significant value in defaulting to "safe" and letting users override to "unsafe but convenient" if they want, which is what fixed versioning would give us.

[...]

PR #723, opened 2023-12-20
Locks down dependencies

KillariDev:
Locks down dependencies so that we are only getting the versions that we intended.
fixes https://github.com/ethereum/solc-js/issues/570


r0qs:
Thanks @KillariDev but for now we decided to not lock the dependencies. So I'm closing this ;)

PR #730, opened 2024-03-05
Update hardhat docker images

r0qs:
Hardhat builds now depends on rust to be installed. So we need to update the docker images similarly to what we did for solidity: https://github.com/ethereum/solidity/commit/e7312a501086e1b0751bc3abbc7e7f08de8b8acd

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,12 @@ commands:
             yarn --version
             pnpm version || >&2 echo "pnpm not installed"
 
+  install-pnpm:
+    steps:
+      - run:
+          name: Install pnpm
+          command: sudo npm install -g pnpm
+
   update-npm:
     steps:
       - run:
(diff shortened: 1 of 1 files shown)


r0qs:
Maybe we should also switch to use hardhat releases instead of their master.

aarlt (approved):
Looks good

cameel:
Looks like CI still fails in `hardhat-core-default-solc`

r0qs:
> Looks like CI still fails in `hardhat-core-default-solc`

Yeah, I "fixed" it now by increasing the nodejs heap allocation like Hardhat does that in their CI: https://github.com/NomicFoundation/hardhat/blob/f33ce2b7bbf92ae94f1755f2c66f2e864370b03b/.github/workflows/hardhat-network-tracing-all-solc-versions.yml#L46. But this error is probably due to some memory leak in their side or related to the nodejs version used.

cameel (approved):

PR #734, opened 2024-03-15
Add a nightly CI run

cameel:
We should really have a nightly run in this repo because there's not that much traffic here and we sometimes miss CI being broken. With a nightly run we won't have to remember to run CI manually and we'll actually notice breakage as it happens, not just before a release.

It would be useful to also have notifications for the Matrix CI channel. And a run with a nightly binary from solc-bin, which would have detected breakage like #732 and, combined with that notification, would let us know when nightlies in solc-bin are broken (https://github.com/ethereum/solidity/issues/13507). I did not do any of those things though. This is just a simple nightly workflow build quickly out of what I already had at hand.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ defaults:
 
 workflows:
   version: 2.1
-  node-multi-build:
+  pr-checks:
     jobs:
       - check-coding-style
       - node-v10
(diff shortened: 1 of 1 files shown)


ekpyron (approved):

cameel:
Not sure what's up with the required `coveralls` job here. It randomly changed name (again) so Github started assuming that the job did not run. And the PR does not even touch anything related to it. On top of that Github did not let me change the requirement to the new name so I thought I'd just remove it and readd after merging, but now the job completely disappeared from settings...

PR #739, opened 2024-05-17
Use hardhat latest releases instead of main branch

r0qs:
Workaround for https://app.circleci.com/pipelines/github/ethereum/solc-js/1540/workflows/d9c4642f-a94e-4aa1-a8e4-90904d689a69/jobs/9433?invite=true#step-113-530140_122. I still need to identify the root cause of the issue (maybe hardhat heuristics?) and report upstream if necessary.

I also imported the `GITHUB_ACCESS_TOKEN` of Solidity CircleCI config to solc-js.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,10 +147,33 @@ commands:
               exit 1
             fi
 
+  fetch-latest-hardhat-release-tag:
+    description: "Uses GitHub API to fetch the latest hardhat release version."
+    steps:
+      - run:
+          name: Retrieve Hardhat latest release tag
+          command: |
+            # Make authenticated requests when the Github token is available
+            if [[ -n "$GITHUB_ACCESS_TOKEN" ]]; then
+              EXTRA_HEADERS=(--header "Authorization: Bearer ${GITHUB_ACCESS_TOKEN}")
+            fi
+            HARDHAT_LATEST_RELEASE_TAG=$(
+              curl \
+                --silent \
+                --location \
+                --fail \
+                --show-error \
+                "${EXTRA_HEADERS[@]}" \
+                https://api.github.com/repos/nomiclabs/hardhat/releases/latest \
+                | jq --raw-output .tag_name \
+            )
+            echo "export HARDHAT_LATEST_RELEASE_TAG='${HARDHAT_LATEST_RELEASE_TAG}'" >> "$BASH_ENV"
+
   provision-hardhat-with-packaged-solcjs:
     description: "Clones Hardhat repository and configures it to use a local clone of solc-js."
     steps:
-      - run: git clone --depth 1 "https://github.com/nomiclabs/hardhat" hardhat/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -150,7 +150,25 @@ commands:
   provision-hardhat-with-packaged-solcjs:
     description: "Clones Hardhat repository and configures it to use a local clone of solc-js."
     steps:
-      - run: git clone --depth 1 "https://github.com/nomiclabs/hardhat" hardhat/
```
cameel, line 153: We also have this: https://github.com/ethereum/solc-js/blob/2f4e680edc2a89ead36d8b1ec4ee637a95111203/.circleci/config.yml#L341

It's already pulling in a release, but there's technically a small risk that their releases on github and npm are not in sync. I would not expect that to happen often but if it happens it would be a hell to debug this.

Overall, it would be better to have a separate task that just fetches the versionn number of the latest release and stores it in a variable. Then all other tasks could just use that number.

cameel (approved):
There's one thing that could be done better, but it's good enough for release so I'm approving. If it passes CI, it's good.

cameel:
Also, it's always good to link the related PRs/issues. In this case the code comes from https://github.com/ethereum/solidity/pull/14875.

.circleci/config.yml
r0qs: Right. But I believe that you cannot access environment variables defined in one job in another job, but we could store the version in the persistent storage and load in another jobs. Otherwise, the option would be to fetch the version at each job, which is what this currently do.

But yeah, the `npm update` would need to install the same version. I will fix that.

cameel: I didn't mean to actually share the value jobs. What I meant was a shared *step*, like for example `show-npm-version`, which the job that needs the value can run and the value gets added to its own env.

Anyway, I wanted to suggest that as a potential improvement, but it should be done separately, if at all. This PR is fine to merge without it.

r0qs: Oh, I just did the change since it was not much different of what the PR was doing anyway.

.circleci/config.yml
```diff
@@ -315,12 +338,15 @@ jobs:
...
           path: boilerplate
+      - fetch-latest-hardhat-release-tag
       - run:
           name: Update to the latest Hardhat release
           command: |
             # We can just use a release here because injection does not require rebuilding it.
             cd boilerplate/
-            npm update hardhat
+            # Install the latest release of Hardhat if the version matches the expected format.
+            [[ "${HARDHAT_LATEST_RELEASE_TAG}" =~ ^hardhat@([0-9]+\.){2}[0-9]+$ ]] && npm install ${HARDHAT_LATEST_RELEASE_TAG}
```
cameel, line 348: ok, this is fine assuming it will fail, when the version does not match and not just continue without installing it.

PR #741, opened 2024-06-10
Update cvc4 to cvc5

r0qs:
Depends on https://github.com/ethereum/solidity/pull/15078

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ return either an error or the result from the solver.  A default
 ``smtchecker.js`` which exports the ``smtCallback`` function that takes 1) a
 function that takes queries and returns the solving result, and 2) a solver
 configuration object. The module ``smtsolver.js`` has a few predefined solver
-configurations, and relies on Z3, Eldarica or CVC4 being installed locally.  It
+configurations, and relies on Z3, Eldarica or cvc5 being installed locally.  It
 exports the list of locally found solvers and a function that invokes a given
 solver.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -152,7 +152,7 @@ return either an error or the result from the solver.  A default
 ``smtchecker.js`` which exports the ``smtCallback`` function that takes 1) a
 function that takes queries and returns the solving result, and 2) a solver
 configuration object. The module ``smtsolver.js`` has a few predefined solver
-configurations, and relies on Z3, Eldarica or CVC4 being installed locally.  It
+configurations, and relies on Z3, Eldarica or CVC5 being installed locally.  It
```
blishko, line 155:
```suggestion
configurations, and relies on Z3, Eldarica or cvc5 being installed locally.  It
```

Just a nitpick, but that's how they want to write the name of the solver now :)

blishko (commented):
Looks simple enough :)

Do we need to test it or have you already done that?

r0qs:
> Looks simple enough :)
> 
> Do we need to test it or have you already done that?

No, I have not done that, but we could indeed add some CI tests. However, even z3 is not currently tested (see: https://github.com/ethereum/solc-js/blob/master/test/smtchecker.ts#L64). Therefore, I'm not sure if it's worth adding them in this PR. We could consider adding tests for all solvers in a subsequent PR.

blishko (approved):
LGTM!

blishko:
OK, let's see if we can add some tests in another PR!

Issue #752, opened 2024-11-13
Replace Hardhat tests with EDR tests in CI


fvictorio:
Hi, this issue is meant to be a continuation of https://github.com/ethereum/solidity/pull/15461 As @cameel pointed out in that PR, we also need to update the CI in this repo to be able to remove the network simulation tests from the Hardhat repository which now live in the [EDR](https://github.com/NomicFoundation/edr) repository.

I was able manage to adapt the CI in the solidity repo without any issues, but here I think I will need some help/guidance.

From what I can tell, there are three Hardhat-related jobs here:
- `hardhat-core-default-solc`
- `hardhat-core-latest-solc`
- `hardhat-sample-project`

The third one, `hardhat-sample-project`, I think can be kept as is, since it's more about Hardhat's compilation pipeline than about network simulation features.

The first one, `hardhat-core-default-solc`, seems a bit weird to me; I'm not sure exactly what it's testing. I don't know if this one can be kept as is or if it can be removed. Adapting it to EDR wouldn't make sense because we don't use npm's solc package there, so the steps here wouldn't do anything.

The second one, `hardhat-core-latest-solc`, definitely needs to be adapted to run in EDR. For this, I guess I'll need to take the `provision-hardhat-with-packaged-solcjs` step and create a new one for EDR. I don't foresee any major inconveniences here though.

So I guess I could use an answer about the first job (keep it or remove it), and to know if I'm missing something in this assessment. Thanks a lot!

cameel:
The point of `hardhat-core-default-solc` and `hardhat-core-latest-solc` was to test solc-js code rather than the compiler binary - the CI in the main repo covers that already. We still needed to use *some* binary though. The most natural thing was to use the binary we package with solc-js, which is what `hardhat-core-latest-solc` does. For some reason that seemed to run only a small subset of your test suite though. `hardhat-core-default-solc` was added to also get the whole suite to run by giving Hardhat the solc version that it uses by default, but still with solc-js from the repo. This is a bit awkward though because we need to replace the binary after installing the package.

In any case, if EDR does not use npm's solc, i.e. solc-js, there's not much point testing it here. I think we still want to run Hardhat's test suite here, at least as long as it keeps using solc-js.

fvictorio:
Thanks for the answer!

[...]

PR #754, opened 2024-12-03
Remove unnecessary Hardhat-related CI jobs

fvictorio:
As discussed in https://github.com/ethereum/solc-js/issues/752, this removes two Hardhat-related jobs that are not necessary and that will stop working soon when we remove those tests from the Hardhat repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,6 @@ workflows:
       - node-current:
           run_coveralls: true
       - build-package
-      - hardhat-core-default-solc: *requires_package
-      - hardhat-core-latest-solc: *requires_package
       - hardhat-sample-project: *requires_package
       - cli-smoke-test: *requires_package
       - solidity-solcjs-ext-test
@@ -35,8 +33,6 @@ workflows:
     jobs:
       - node-current
       - build-package
-      - hardhat-core-default-solc: *requires_package
-      - hardhat-core-latest-solc: *requires_package
       - hardhat-sample-project: *requires_package
       - cli-smoke-test: *requires_package
       - solidity-solcjs-ext-test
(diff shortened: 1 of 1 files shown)


fvictorio:
The second commit fixes an unrelated thing in the CI: it assumes that the Hardhat repo only has releases related to Hardhat, but we also publish our plugins releases there. This means that if the latest release is a plugin, the job fails.

I fixed that by getting all releases, and getting the first one whose tag starts with `hardhat@`.

.circleci/config.yml
```diff
@@ -164,25 +154,11 @@ commands:
                 --fail \
                 --show-error \
                 "${EXTRA_HEADERS[@]}" \
-                https://api.github.com/repos/nomiclabs/hardhat/releases/latest \
-                | jq --raw-output .tag_name \
+                https://api.github.com/repos/nomiclabs/hardhat/releases \
+                | jq --raw-output 'map(select(.tag_name | test("^hardhat@"))) | .[0].tag_name' \
```
r0qs, line 158: Is the latest release always the first element in the list?

r0qs (commented):
Thank you very much @fvictorio! It looks good to me, except for one small comment :)

r0qs:
> The node-v10 job is still failing, but that also seems unrelated to my changes here.

Yeah, it is unrelated. I will fix it in a follow up PR, thanks.

.circleci/config.yml
fvictorio: It is according to my single, manual test :sweat_smile: Sadly, [the docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) don't specify this.

But I think even in the (IMO unlikely) case that you get an older release, this is still better than failing every time the latest release is a Hardhat plugin, something that happens often enough.

r0qs (approved):

r0qs:
@cameel this cannot be merged because `ci/circleci: hardhat-core-latest-solc` is marked as required and I don't have access to the repo settings to change that. Note that this job was removed by this PR ;)

fvictorio:
That job must be failing because I removed those tests from the Hardhat repo thinking this PR was already merged, sorry about that!

cameel (approved):

cameel:
Required jobs updated.

PR #756, opened 2025-01-23
Drop nodejs 10

r0qs:
This PR removes Node.js 10 from our CI as it started failing due to many dependencies increasing their Node.js version requirements. See: https://app.circleci.com/pipelines/github/ethereum/solc-js/1775/workflows/14eb5b72-7624-47e5-a302-a407c1fd4f58/jobs/11376/parallel-runs/0?invite=true#step-105-0_136 and this error: https://app.circleci.com/pipelines/github/ethereum/solc-js/1775/workflows/14eb5b72-7624-47e5-a302-a407c1fd4f58/jobs/11376?invite=true#step-108-209_118, which is because `Object.fromEntries` is only available in Node.js 12+. We could potentially workaround the issue using `DISABLE_V8_COMPILE_CACHE=1`, but I believe that it makes more sense to update the packages.

Also, ESLint v8 dropped support for Node.js 10, as noted here https://github.com/eslint/eslint/issues/15560#issuecomment-1055090605.

As a result, we should probably consider bumping everything to align with the newer Node.js versions. And we might need to revisit the suggestion from https://github.com/ethereum/solc-js/pull/723 to lock our dependencies for better consistency.

Note, I decided to bump some versions, but not the major ones yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "compiler"
   ],
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=12.0.0"
   },
   "files": [
     "common/*.js",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,6 @@ workflows:
   pr-checks:
     jobs:
       - check-coding-style
-      - node-v10
       - node-v12
       - node-v14
       - node-v16
@@ -345,10 +344,6 @@ jobs:
       - run: cd solidity/ && curl "https://binaries.soliditylang.org/bin/soljson-nightly.js" --location --output soljson.js
       - run: cd solidity/ && test/externalTests/solc-js/solc-js.sh "$(realpath soljson.js)" "$(scripts/get_version.sh)" "$(realpath ../solc-js/)"
 
-  node-v10:
-    <<: *node-base
-    docker:
-      - image: cimg/node:10.24
   node-v12:
     <<: *node-base
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ workflows:
       - node-v16
       - node-v18
       - node-v20
+      - node-v22
       - node-current:
           run_coveralls: true
       - build-package
@@ -364,6 +365,10 @@ jobs:
     <<: *node-base
     docker:
       - image: cimg/node:20.9
+  node-v22:
+    <<: *node-base
+    docker:
+      - image: cimg/node:22.14
   node-current:
     <<: *node-base
     docker:


cameel (approved):
Ok, let's do it. Node 10 is ancient at this point and I don't think it's essential. We try to support even very old versions mostly because it's not much effort for us. But if the broader ecosystem is dropping it, it's time to let it go.

Please remember to add a changelog entry in the main repo, [like we did when we dropped nodejs v8](https://github.com/ethereum/solidity/blob/v0.8.28/Changelog.md?plain=1#L464-L466). The changelog for solc-js is maintained there.

cameel:
> As a result, we should probably consider bumping everything to align with the newer Node.js versions.

That's not how it works. We don't require specific versions of dependencies, we support whole ranges and it's the downstream app that chooses a single version from that range. On our end generally shouldn't have to bump anything as long as `package.json` is written correctly. Version range should be open-ended, allowing new versions as long as they're compatible according to semver.

What we may need to do from time to time is to update to breaking versions, but that's not a simple version bump. It will usually require changes in code, otherwise the version would not be breaking. It may sometime happen that the breaking changes do not affect our usage, but we need to verify that. And it just means that we can broaden the version range without dropping support for the older one.

BTW, support for older versions is something we should be testing. Our tests run with latest versions of dependencies, but we should also have one with oldest allowed ones to notice when we introduce something that does not work with them.

> Note, I decided to bump some versions, but not the major ones yet.

Feel free to do that update in a separate PR. Just make sure we're not restricting them too much. Like, we don't want to suddenly reject older versions that are still commonly used by frameworks. Ideally, we'd just add mark the newer version as supported without actually dropping the older one (as long as it still works).

>  And we might need to revisit the suggestion from #723 to lock our dependencies for better consistency.

How would a `package-lock.json` help with anything? It's just a snapshot of dependency versions at a random point in time. It also does not mean anything to us, because we still support the whole range. IMO testing latest and oldest combinations of versions would be much more helpful.

package.json
cameel, line 1: Actually, do we even need to bump these? They all use `^` so doing that is not necessary to get newer versions. Instead, what we're doing is just to preventing the older ones from being installed. Are these older versions known to be broken?

cameel, line 1: By the way, you should also update this bit:
```json
  "engines": {
    "node": ">=10.0.0"
  },
```

cameel:
BTW, I see that the latest node.js version is now 23 so we should add a run on 22.

r0qs:
> > As a result, we should probably consider bumping everything to align with the newer Node.js versions.
> 
> That's not how it works. We don't require specific versions of dependencies, we support whole ranges and it's the downstream app that chooses a single version from that range. On our end generally shouldn't have to bump anything as long as `package.json` is written correctly. Version range should be open-ended, allowing new versions as long as they're compatible according to semver.
> 
> What we may need to do from time to time is to update to breaking versions, but that's not a simple version bump. It will usually require changes in code, otherwise the version would not be breaking. It may sometimes happen that the breaking changes do not affect our usage, but we still need to verify that it's the case. And it just means that we can broaden the version range without dropping support for the older one.
> 
> BTW, support for older versions is something we should be testing. Our tests run with latest versions of dependencies, but we should also have one with oldest allowed ones to notice when we're introducing something that does not work with them.

Ah, got it. Yeah, it indeed does make sense to support a range of dependencies.

> > Note, I decided to bump some versions, but not the major ones yet.
> 
> Feel free to do that update in a separate PR. Just make sure we're not restricting them too much. Like, we don't necessarily want to update if frameworks are still mostly on an older, incompatible version. Whenever possible and we can work with both old and new version, just add mark the newer version as supported without actually dropping the older one.
> 
> > And we might need to revisit the suggestion from #723 to lock our dependencies for better consistency.
> 
> How would a `package-lock.json` help with anything? It's just a snapshot of dependency versions at a random point in time. It also does not mean anything to us, because we still support the whole range from `package.json`, not just the set from the lock file. IMO testing latest and oldest combinations of versions would be much more helpful.
> 
> Or do you mean that we should actually restrict the library to one specific version of each dependency? I don't think that's a good idea.

My original comment was motivated by concerns raised in https://github.com/ethereum/solc-js/issues/570, rather than by the specific versions set in https://github.com/ethereum/solc-js/pull/723. However, after revisiting the issue, I agree with you, and it does seem preferable to keep things up to date and support a range of versions rather than fixing them. Should we then consider this issue and PR closed?

package.json
r0qs: Indeed. I removed the bump as you suggested. And let's do it in the way you have done in the other PR.

cameel (approved):

cameel:
Well, I don't think we should merge #723 and it's outdated by now so it's probably best to close it.

As for #570, maybe it's better to still leave it open. I don't think having a lock file in a library is a good idea, but in the issue it was basically me vs one other opinion so I was waiting for more discussion to happen there. I know that many people insist on always using that file, but so far I have not seen a compelling reason. See for example [When not to use package-lock.json](https://dev.to/gajus/stop-using-package-lock-json-or-yarn-lock-3ddi) and people's comments there. So far no one else said anything though.

PR #757, opened 2025-01-24
Add test runs with minimum dependency versions

cameel:
Depends on #756.


.circleci/config.yml
```diff
@@ -8,14 +8,19 @@ workflows:
...
+          min_dependencies: true
       - node-v14
       - node-v16
       - node-v18
       - node-v20
       - node-current:
           run_coveralls: true
+      - node-current:
+          name: node-current-min-dependencies
+          min_dependencies: true
```
cameel, line 23: For some reason CircleCI decided to change the name of the existing job to `node-current-1`. Not sure why, this does not happen with `node-v12` for example. I tried adding `name: node-current` but it does not change this.

I guess we'll have to accept this and update the name of the required job in settings if we decide to merge the PR. I think we should make the `min-dependencies` jobs required as well.

.circleci/package-json-with-min-dependencies.sh
cameel, line 1: I wish npm had built-in support for this but apparently it does not so it has to be done on our own. I saw an old feature request for `npm install --minimum`, but it was eventually closed as stale.

.circleci/config.yml
cameel: No. It has to be `node-current` for both because that's the name of the job that's being invoked in both cases: https://github.com/ethereum/solc-js/blob/4aec1e70623a481b32dbe54f248084e90ed93b2f/.circleci/config.yml#L457-L460

The name of the run normally comes from that invoked job, but if it's invoked more than once, CircleCI adds numbers to disambiguate it. That's why there's the `name` parameter - note that it does not come from the job definition (see `node-base`). It's a built-in CircleCI thing meant for providing a custom name.

See also [`t_ext` in Solidity CI config](https://github.com/ethereum/solidity/blob/v0.8.28/.circleci/config.yml#L1864-L1881). We do it there the same way and it works.

PR #758, opened 2025-01-25
Support broader range of dependency versions

cameel:
Depends on #757.
Replaces the version bump attempted in #756.

This PR extends the ranges of dependency versions marked as supported both ways:
- Where possible I lowered the minimum. The new versions are the lowest ones that install, build and pass tests on all our supported node.js versions. There could still be some more subtle breakage not detected by tests but the same can be said for current versions - it's not like we chose them all that carefully. They're mostly whatever was the latest when we last bumped them. I checked history for the few cases where the bumps were clearly intentional and retained them.
- I also relaxed upper limits so that we accept newer versions even across breaking upgrades. Most packages seem to be actually pretty widely compatible. There were only a few that weren't and I added restrictions only for those.


package.json
```diff
@@ -46,32 +46,32 @@
...
   "homepage": "https://github.com/ethereum/solc-js#readme",
   "dependencies": {
-    "command-exists": "^1.2.8",
-    "commander": "^8.1.0",
-    "follow-redirects": "^1.12.1",
-    "js-sha3": "0.8.0",
-    "memorystream": "^0.3.1",
-    "semver": "^5.5.0",
-    "tmp": "0.0.33"
+    "command-exists": ">=1.1.0",
```
cameel, line 49: Note that I replaced `^` with `>=` in all version ranges. I'm still on the fence whether that's a good idea or whether I should always put a `<` there to keep major version updates manual:

Pros of `<=`:
- We almost never update these versions and this forces projects using `solc-js` to use old versions too (or do hacky workarounds). `<=` forces the update on us by default and we can always restrict with `<` as needed.
- There's not much risk of this seriously breaking a downstream app that's already using the package. If it happens, the app can always restrict the dependency in its own `package.json`.

Cons of `<=`:
- Already released versions in `npm` will stay unlimited forever and eventually stop working with latest versions of dependencies because one of them will inevitably introduce an actual breaking change. Someone trying to add a dependency on an older version of solc-js may be confused by this.
- There's some risk of a new breaking version appearing during our release process. Though in solc-js this would be easy to fix quickly, because CI is quick and it gets tagged late in the process.

package.json
```diff
@@ -46,32 +46,32 @@
...
+    "commander": ">=8.0.0 <12.0.0",
```
cameel, line 50: Versions `>=13.0.0` does not pass tests. We could probably support it if we wanted but it would require adjustments in the code.

Versions `>=12.0.0` trigger failures on older node.js versions due to the use of `??` operator or its variants.

package.json
```diff
@@ -46,32 +46,32 @@
...
+    "eslint": ">=7.12.1 <8.0.0",
+    "eslint-config-standard": ">=16.0.3 <17.0.0",
+    "eslint-plugin-import": ">=2.25.0",
+    "eslint-plugin-node": ">=11.1.0",
+    "eslint-plugin-promise": ">=5.0.0 <7.0.0",
+    "nyc": ">=15.0.0",
+    "tape": ">=4.11.0",
+    "tape-spawn": ">=1.0.0",
+    "ts-node": ">=10.0.0",
+    "typescript": ">=4.2.2 <5.0.0"
```
cameel, line 74: Typescript, eslint and their plugins here are restricted only because they have issues on node.js 12 and 14. There are failures due to the use of `??` operator or its variants. On the other hand there are no issues on the latest node.js.

Not sure how to handle this. I guess node.js 12 and 14 are past EOL now so it may be ok to just remove the restriction. The issue is that our tests fail, so we would have to patch `package.json` in our tests.

Another possibility might be to just drop official v12 and v14 support.

PR #760, opened 2025-01-28
Update LICENSE

maximevtush:
Updated the copyright year from 2016 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2025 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 
+Copyright (c) 2025
```
cameel, line 3: This is not necessary. The year here is the year of "the first publication of the work". It does not need to be bumped constantly.

PR #789, opened 2026-03-21
Fetch Hardhat v2 version from npm instead of GitHub API

r0qs:
The GitHub API `/releases` endpoint returns 30 results by default. Hardhat v3 releases have pushed all v2 releases off the first page, so the existing jq filter returns `null`. This can be reproduced with:

```bash
curl -sL https://api.github.com/repos/nomiclabs/hardhat/releases | jq 'map(select(.tag_name | test("^hardhat@2"))) | length'
# returns: 0
```

Also see the failing job: https://app.circleci.com/jobs/github/argotorg/solc-js/15386

This PR replaces GitHub API release lookup with `npm view` to fetch the latest v2 version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,29 +139,20 @@ commands:
               exit 1
             fi
 
-  fetch-latest-hardhat-release-tag:
-    description: "Uses GitHub API to fetch the latest hardhat release version."
+  fetch-latest-hardhat-v2-version:
+    description: "Uses npm registry to fetch the latest hardhat v2 release version."
     steps:
       - run:
-          name: Retrieve Hardhat latest release tag
+          name: Retrieve latest Hardhat v2 version
           command: |
-            # Make authenticated requests when the Github token is available
-            if [[ -n "$GITHUB_ACCESS_TOKEN" ]]; then
-              EXTRA_HEADERS=(--header "Authorization: Bearer ${GITHUB_ACCESS_TOKEN}")
-            fi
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -139,29 +139,20 @@ commands:
...
-                --silent \
-                --location \
-                --fail \
-                --show-error \
-                "${EXTRA_HEADERS[@]}" \
-                https://api.github.com/repos/nomiclabs/hardhat/releases \
-                | jq --raw-output 'map(select(.prerelease == false and (.tag_name | test("^hardhat@2")))) | .[0].tag_name' \
+            HARDHAT_V2_VERSION=$(
+              npm view hardhat@'>=2.0.0 <3.0.0' version --json \
+                | jq --raw-output '.[-1]' \
```
nikola-matic:
```suggestion
                | jq --raw-output 'last' \
```

A nit, but I like this better :)

.circleci/config.yml
```diff
@@ -257,14 +248,19 @@ jobs:
       - install-dependencies:
           cache-id: hardhat-hackathon-boilerplate
           path: boilerplate
-      - fetch-latest-hardhat-release-tag
+      - fetch-latest-hardhat-v2-version
       - run:
           name: Update to the latest Hardhat release
```
nikola-matic:
```suggestion
          name: Update to the latest Hardhat v2 release
```

nikola-matic (commented):
Nice, much better way of doing this.

nikola-matic (approved):

PR #795, opened 2026-07-21
Fix EAGAIN crash when reading --standard-json input from a pipe

### LINKED ISSUE #460: "Error: EAGAIN: resource temporarily unavailable, read" when using --standard-json flag
When trying to use the command line with the flag **--standard-json**, the following error is thrown

```
solcjs --standard-json Example.sol
```
```
internal/fs/utils.js:230
    throw err;
    ^

Error: EAGAIN: resource temporarily unavailable, read
    at Object.readSync (fs.js:537:3)
    at tryReadSync (fs.js:352:20)
    at Object.readFileSync (fs.js:389:19)
    at Object.<anonymous> (/usr/lib/node_modules/solc/solcjs:37:18)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  errno: -11,
  syscall: 'read',
  code: 'EAGAIN'
}
```

cavdarahmet:
Fixes #460

`fs.readFileSync(process.stdin.fd)` can throw `EAGAIN` when stdin is a pipe
that has no data available yet at the moment of the (synchronous) read (see
https://github.com/nodejs/node/issues/7439 for the underlying node.js
limitation). This is not just theoretical: it's been reported multiple times
over the years (2020, 2021, 2024) and is still reproducible today on Node.js
v24.

Fix: read the input via the stdin stream's `data`/`end` events instead, which
handles this correctly regardless of how the data arrives on the pipe.

Added a regression test that reproduces the original crash by writing the
standard-json input one byte at a time (a single atomic write does not
trigger the race).

### DIFF
--- a/solc.ts
+++ b/solc.ts
@@ -123,136 +123,153 @@ function reformatJsonIfRequested (inputJson) {
 let callbacks;
 if (options.basePath || !options.standardJson) { callbacks = { import: readFileCallback }; }
 
-if (options.standardJson) {
-  const input = fs.readFileSync(process.stdin.fd).toString('utf8');
-  if (program.verbose) { console.log('>>> Compiling:\n' + reformatJsonIfRequested(input) + '\n'); }
-  let output = reformatJsonIfRequested(solc.compile(input, callbacks));
-
-  try {
-    if (smtsolver.availableSolvers.length === 0) {
-      console.log('>>> Cannot retry compilation with SMT because there are no SMT solvers available.');
-    } else {
-      const inputJSON = smtchecker.handleSMTQueries(JSON.parse(input), JSON.parse(output), smtsolver.smtSolver, smtsolver.availableSolvers[0]);
-      if (inputJSON) {
-        if (program.verbose) { console.log('>>> Retrying compilation with SMT:\n' + toFormattedJson(inputJSON) + '\n'); }
-        output = reformatJsonIfRequested(solc.compile(JSON.stringify(inputJSON), callbacks));
-      }
-    }
-  } catch (e) {
-    const addError = {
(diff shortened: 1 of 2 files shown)


cavdarahmet:
The red `node-v24-1` check is the Coveralls upload, not the tests.

That job runs the suite twice: the `test` step passes, and the `coveralls` step runs it again
under nyc and also passes — 1165/1165, including the new `standard json with input arriving on
stdin in small chunks` case. It then fails when uploading the report:

    Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}

This looks specific to builds from forks rather than to this PR. The same `node-v24-1` job runs
the coveralls step and succeeds on in-repo branches — build 16419 on master, and 16408 on
`smtchecker_ignore_bmc_deprecation_warning`, which was itself a PR. Meanwhile the only two fork
builds that reached that job in the last ~1000 CI runs, this one (16374) and #791 (15928),
both failed on it with the identical 422, two months apart. That is consistent with CircleCI
not passing the Coveralls token to forked PRs by default, which would leave the upload unable
to identify the repository.

Nothing in this PR can change that, and it will affect any external contribution the same way;
gating `run_coveralls` on the token being present would keep the check green for forks.

The branch still merges cleanly against master and is ready for review whenever you have a moment.

